PR #24, opened 2016-03-08
WIP: Test dplyr as part of the overall tests

krlmlr:
Currently requires hadley/testthat#404.

A few tests fail (but again succeed locally); need to investigate.

@hadley: Is this an approach we want to follow? For this case it would help to have a protected "production" branch in dplyr where only successful builds are accepted.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,6 @@ env:
 r_github_packages:
   - hadley/devtools
   - hadley/dplyr
-  - hadley/testthat#404
 
 after_success:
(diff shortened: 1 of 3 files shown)


hadley:
Not sure why you want to do this?

krlmlr:
I think the dplyr tests are more comprehensive than the tests here. If not, there's a lot of duplication in the tests.

This is until there's a dedicated testing package (dplyrtest?).

krlmlr:
Doesn't work anymore after extraction of dbplyr.

PR #42, opened 2017-04-21
Compat dplyr 0.6.0

lionel-:


lionel-:
@hadley pinging you because I can't request a review here. Also could you make me a collaborator please?

lionel-:
The problem with the current load-time registration is that it seems harder to maintain because the registration does not happen near the method definition like with roxy tags. Would this UI work?

```r
filter.data.table <- function(.data, ...) {
  filter_(.data, .dots = lazyeval::lazy_dots(...))
}
mtd_export("filter", "data.table", dep_package("dplyr", `>`, "0.5.0"))
```

With the following tools to add methods to a growing pairlist (it seemed cleaner to take a pairlist as data structure because we can't allocate the whole vector in advance).

```r

.onLoad <- function(libname, pkgname) {
  if (exists(".s3_methods", inherits = TRUE)) {
    # TODO: Use pairlist walker with rlang tools
    lapply(.s3_methods, mtd_register)
  }

  invisible()
}

mtd_register <- function(mtd) {
  if (!deps_qualifies(mtd$deps)) {
    return(NULL)
  }

  gric <- mtd$gric
  class <- mtd$class
  pkg <- mtd$pkg

  stopifnot(is.character(pkg), length(pkg) == 1)
  stopifnot(is.character(gric), length(gric) == 1)
  stopifnot(is.character(class), length(class) == 1)

  envir <- asNamespace(pkg)
  fun <- get(paste0(gric, ".", class), envir = parent.frame())
  stopifnot(is.function(fun))


  if (pkg %in% loadedNamespaces()) {
    registerS3method(gric, class, fun, envir = envir)
  }

  # Always register hook in case package is later unloaded & reloaded
  setHook(
    packageEvent(pkg, "onLoad"),
    function(...) {
      registerS3method(gric, class, fun, envir = envir)
    }
  )
}

mtd_export <- function(gric, class, ..., ns = NULL) {
  method_ns <- parent.frame()

  if (is.null(ns)) {
    gric_fn <- get(gric, envir = method_ns, mode = "function")
    ns <- environment(gric_fn)
  }
  stopifnot(isNamespace(ns))

  if (!exists(".s3_methods", envir = method_ns)) {
    assign(".s3_methods", pairlist(), envir = method_ns)
  }

  # TODO: Use rlang pairlist helpers to grow method list
  mtd <- mtd(gric, class, ns, ...)
  .s3_methods <<- c(pairlist(mtd), .s3_methods)
}

mtd <- function(gric, class, pkg, ...) {
  deps <- list(...)
  stopifnot(all(vapply(deps, inherits, logical(1), "dep_pkg")))
  list(gric = gric, class = class, pkg = pkg, deps = deps)
}

dep_package <- function(pkg, cmp, version) {
  structure(list(pkg, cmp, version), class = c("dep_pkg", "dep"))
}

deps_qualifies <- function(deps) {
  if (is_null(deps)) return(TRUE)

  for (dep in deps) {
    if (!dep[[2]](utils::packageVersion(dep[[1]]), dep[[3]])) {
      return(FALSE)
    }
  }

  TRUE
}
```

hadley:
I think we should do manually in `.onLoad()` for now, but file this idea into roxygen2 so we can provide something like `@registerS3dynamic`

lionel-:
@hadley then I think we should continue using `@export`. Otherwise it's a lot of copy/pasting and opportunity for something going wrong.

hadley:
I don't trust the namespace switch approach right now. Let's just do the hacky onload registration for now because it's at least consistent within one package. We can talk more about the reasoning when we next meet.

Issue #21, opened 2016-03-08
Avoid copying data tables during a join


hadley:
Moved from https://github.com/hadley/dplyr/issues/659, by @matthieugomez

Issue #13, opened 2016-03-08
auto_copy should distinguish non-tbl data.tables from data.frames


hadley:
Moved from https://github.com/hadley/dplyr/issues/1224, by @richierocks

If you try to join a non-tbl data table to a data frame with `copy = TRUE`, you get the following error:

```
library(dplyr)
library(data.table)

DT <- data.table(id = 1:10, x = runif(10))
DF <- data.frame(id = sample(10, 10, replace = TRUE), y = letters[1:10])

left_join(DT, DF, copy = TRUE)
## Joining by: "id"
## Error in data.table::setkeyv(y, by$x) : x is not a data.table
```

This works as expected:

```
left_join(as.tbl(DT), DF, copy = TRUE)
```

Closes #14 because of https://github.com/hadley/dtplyr/pull/45#discussion_r115852225.

hadley:
Seems to be ok now:

``` r
library(data.table)
library(dtplyr)
library(dplyr, warn.conflicts = FALSE)

dt <- data.table(id = 1:10, x = runif(10))
df <- data.frame(id = sample(10, 10, replace = TRUE), y = letters[1:10])

class(left_join(dt, df, copy = TRUE))
#> Joining, by = "id"
#> [1] "tbl_dt"     "tbl"        "data.table" "data.frame"
```

<sup>Created on 2019-06-13 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

PR #45, opened 2017-05-10
Join using merge

### LINKED ISSUE #14: Joins need to check that `y` is a data.table
Moved from https://github.com/hadley/dplyr/issues/1217, by @helix123 

``` R
library(dplyr)
library(dplyr)

dt <- data.table::data.table(x = 1:3, y = 3:1)
df <- data.frame(x = 3:1, z = 1:3)

anti_join(dt, df)
left_join(dt, df)
```

### LINKED ISSUE #20: Join with different keys
Since `data.table::merge()` now has `by.x` and `by.y` arguments.

Moved from https://github.com/hadley/dplyr/issues/694, by @cderv

christophsax:
In data.table v1.9.6, `merge.data.table` has gained a `by.x` and `by.y` argument. Also it does not copy anymore, and does not need a key, so using `merge` to perform joins should be more efficient and more convenient.

From `?merge.data.table`:

> In versions <= v1.9.4, if the specified columns in by was not the key (or head of the key) of x or y, then a copy is first rekeyed prior to performing the merge. This was less performant and memory inefficient. The concept of secondary keys (implemented in v1.9.4) was used to overcome this limitation from v1.9.6+. No deep copies are made anymore and therefore very performant and memory efficient. Also there is better control for providing the columns to merge on with the help of newly implemented by.x and by.y arguments.

Using column renaming and the `on =` argument for non-`merge` joins, `semi_join.data.table` and `anti_join.data.table`, saves the need for `setkey` and copying.

This should fix #20, #21.

@krlmlr, we checked this out together, you may have another look. I had to remove some `with = FALSE`, which only work for the `j`, not for the `i` argument. But also without `with = FALSE`, the current solution works if a column is named `y_labels`, so we should be fine. Also `nomatch = 0L` cannot be used in an anti join.


R/joins.R
```diff
@@ -35,15 +35,6 @@ join_dt <- function(op) {
   # nocov start
   template <- substitute(function(x, y, by = NULL, copy = FALSE, ...) {
     by <- dplyr::common_by(by, x, y)
-    if (!identical(by$x, by$y)) {
-      stop("Data table joins must be on same key", call. = FALSE)
-    }
-    y <- dplyr::auto_copy(x, y, copy = copy)
```
krlmlr: We might want to keep this `auto_copy()` call.

R/joins.R
```diff
@@ -55,25 +46,36 @@ join_dt <- function(op) {
...
-right_join.data.table  <- join_dt(merge(x, y, by = by$x, all.y = TRUE, allow.cartesian = TRUE))
+right_join.data.table <- join_dt({merge(x, y, by.x = by$x, by.y = by$y, all.y = TRUE, allow.cartesian = TRUE)})
 
 #' @rdname join.tbl_dt
 semi_join.data.table  <- join_dt({
-  # http://stackoverflow.com/questions/18969420/perform-a-semi-join-with-data-table
-  w <- unique(x[y, which = TRUE, allow.cartesian = TRUE])
-  w <- w[!is.na(w)]
-  x[w]
+   y <- as.data.table(y)
```
krlmlr: `auto_copy()` should take care of that.

krlmlr: Perhaps we could refactor the common parts in a function?

christophsax: Without `copy = TRUE` we still would get:

```
> library(dplyr)
library(dplyr)
dt <- data.table::data.table(x = 1:3, y = 3:1)
df <- data.frame(x = 3:1, z = 1:3)
anti_join(dt, df)
Joining, by = "x"
Error in `[.data.frame`(y, , by_y, with = FALSE) : 
  unused argument (with = FALSE)
```

This could be avoided with `y <- as.data.table(y)`

christophsax:
Closed the two other PR and will open again if this is merged.

R/joins.R
christophsax: Perhaps with separate templates for merge and non-merge joins? As [here](https://github.com/hadley/dtplyr/compare/master...christophsax:separate-template-merge-non-merge)?

For the `suffix` argument, we need to treat them differently anyway.

krlmlr: But wouldn't #13 also take care of that case? I think it's better to have automatic coercion from data frame to data table under our control, and perhaps give a message.

christophsax: Fair enough, will remove it.

krlmlr: Actually, I was thinking about a simple function that computes `filter_y`.

Templating: We could also do something like

```
join_dt <- function(op, suffix = TRUE) {
  if (suffix) suffix_op <- ...
  else suffix_op <- NULL
  template <- substitute({
    ...
    suffix_op
    ...
    op
    ...
  })
  f <- ...
  if (!suffix) formals(f) <- formals(f)[names(formals(f)) != "suffix"]
  ...
}
```

Both solutions are feasible, but really look too complicated for the task at hand. In the end it seems simpler to admin some duplication and get rid of the templating mechanism. But this doesn't have to be in that PR.

krlmlr: Scratch that, `suffix_op` affects `op`. The conclusion still holds -- let's expand the templates.

christophsax: That means #14 is not yet fixed but will only be if #13 is done.

krlmlr: I think that's reasonable.

krlmlr: My last comment may have been ambiguous, sorry for that. I was thinking about expanding the templates to six simple functions that have a few lines of code in common.

christophsax: ok, done, I also like it better that way. Also added a small test.

R/joins.R
```diff
@@ -31,49 +31,61 @@
...
+  y_filter <- y[, by_y, with = FALSE]
+  names(y_filter) <- by_x
+  w <- x[!y_filter, which = TRUE, on = by_x]
+  out <- x[sort(unique(w))]
+  grouped_dt(out, groups(x))
+}
 
 #' @rdname join.tbl_dt
-anti_join.data.table <- join_dt({x[!y, allow.cartesian = TRUE]})
+full_join.data.table <- function(x, y, by = NULL, copy = FALSE, ...){
```
krlmlr: Can you please move the full join above the semi join? I think this is the order we use elsewhere, too.

R/joins.R
```diff
@@ -31,49 +31,61 @@
...
+}
 
 #' @rdname join.tbl_dt
-right_join.data.table  <- join_dt(merge(x, y, by = by$x, all.y = TRUE, allow.cartesian = TRUE))
+semi_join.data.table <- function(x, y, by = NULL, copy = FALSE, ...) {
+  by <- dplyr::common_by(by, x, y)
+  y <- dplyr::auto_copy(x, y, copy = copy)
+  by_x <- by$x
+  by_y <- by$y
+  y_filter <- y[, by_y, with = FALSE]
```
krlmlr: Is the first comma still necessary with data.table?

R/joins.R
```diff
@@ -31,49 +31,61 @@
...
+  names(y_filter) <- by_x
```
krlmlr: Perhaps `y_trimmed <- trim_y_for_semi_join(y, by)`, or even a better verb than "trim", instead of `y_filter`? This removes duplication and adds readability IMO.

R/joins.R
christophsax: Without it, it would take it for the `i`, not the `j` argument. So I guess, yes.

R/joins.R
```diff
@@ -31,49 +31,66 @@
...
+}
 
 #' @rdname join.tbl_dt
-# http://stackoverflow.com/a/15170956/946850
-full_join.data.table <- join_dt({merge(x, y, by = by$x, all = TRUE, allow.cartesian = TRUE)})
+anti_join.data.table <- function(x, y, by = NULL, copy = FALSE, ...) {
+  by <- dplyr::common_by(by, x, y)
+  y <- dplyr::auto_copy(x, y, copy = copy)
+  by_x <- by$x
+  by_x <- by$x
```
krlmlr: Can you please double-check, also if we could perhaps write `by$x` instead? The docs for the `on` argument suggest that we could also avoid renaming the columns in `y` altogether:

```r
on <- rlang::set_names(by$y, by$x)
w <- x[!y, which = TRUE, on = on]
...
```

Would that work with data.table 1.9.6?

dplyr already imports rlang, so no extra dependency added, but we need to explicitly declare the import in DESCRIPTION to be able to use it here.

lionel-: you can also use `@import rlang` and call rlang functions unqualified. I have started on a tidyevalish implementation (not fully tidy because we'll need to flatten quosures) a couple weeks ago, so we'll import it in the future anyway.

lionel-:
> Thanks, looks much simpler now without the templates.

Agreed. Note that in cases where it makes sense, you can now do:

```r
expr <- quote(a <- b)
exprs <- list(quote(foo(a)), quote(foo(b)))

fn <- expr_interp(function(...) {
  !! expr
  !!! exprs
})
```

R/joins.R
christophsax: Yes, indeed. I wasn't aware of that. And it was already introduced in data.table 1.9.6.

Since using `by$y` in `trim_y_for_semi_join` can be used as well, I removed the one line function altogether.

Also added `@import rlang` and called `set_names` unqualified.

krlmlr:
Thanks!

PR #47, opened 2017-05-10
Tests

christophsax:
Some tests to check if the the data.table joins work the same way as the dplyr joins.


krlmlr:
#24 contains an approach to run dplyr's tests for dtplyr, but I'm not sure if it still works.

Issue #51, opened 2017-06-17
Error in .subset(x, j) : invalid subscript type 'language' when using mutate_all


adamryczkowski:
```
library(data.table)
library(dtplyr)
library(dplyr)
dt <- data.table(a=1:10, b=1:10, c=1:10)
dt %>%  mutate_all(as.integer)
# Error in .subset(x, j) : invalid subscript type 'language'
```

When we remove data.table nature of the dt, everything works:

```
library(data.table)
library(dtplyr)
library(dplyr)
dt <- data.table(a=1:10, b=1:10, c=1:10)
df <- as.data.frame(dt)
df %>% mutate_all(as.integer)

#     a  b  c
# 1   1  1  1
# 2   2  2  2
# 3   3  3  3
# 4   4  4  4
# 5   5  5  5
# 6   6  6  6
# 7   7  7  7
# 8   8  8  8
# 9   9  9  9
# 10 10 10 10
```

```
> sessionInfo()
R version 3.4.0 (2017-04-21)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 16.04.2 LTS

Matrix products: default
BLAS: /usr/lib/libblas/libblas.so.3.6.0
LAPACK: /usr/lib/lapack/liblapack.so.3.6.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=pl_PL.UTF-8        LC_COLLATE=en_US.UTF-8     LC_MONETARY=pl_PL.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=pl_PL.UTF-8       LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=pl_PL.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] bindrcpp_0.1      data.table_1.10.4 dplyr_0.7.0       dtplyr_0.0.2.9000

loaded via a namespace (and not attached):
 [1] Rcpp_0.12.11     digest_0.6.12    withr_1.0.2      assertthat_0.2.0 R6_2.2.1         git2r_0.18.0     magrittr_1.5     httr_1.2.1       rlang_0.1.1     
[...]

Issue #43, opened 2017-05-07
Scoped *_all/*_at/*_if functions break on tbl_dt


alistaire47:
I ran into an issue today when trying to use `summarise_all` on a data.table:

``` r
library(dplyr)
library(data.table)
library(dtplyr)

dtcars <- as.data.table(mtcars)

# what should happen
mtcars %>% group_by(cyl) %>% summarise_all(mean)
#> # A tibble: 3 × 11
#>     cyl      mpg     disp        hp     drat       wt     qsec        vs
#>   <dbl>    <dbl>    <dbl>     <dbl>    <dbl>    <dbl>    <dbl>     <dbl>
#> 1     4 26.66364 105.1364  82.63636 4.070909 2.285727 19.13727 0.9090909
#> 2     6 19.74286 183.3143 122.28571 3.585714 3.117143 17.97714 0.5714286
#> 3     8 15.10000 353.1000 209.21429 3.229286 3.999214 16.77214 0.0000000
#> # ... with 3 more variables: am <dbl>, gear <dbl>, carb <dbl>

# what does happen
dtcars %>% group_by(cyl) %>% summarise_all(mean)
#> Error in !funs: invalid argument type

# explicitly coercing to tbl_dt doesn't help
dtcars %>% tbl_dt() %>% group_by(cyl) %>% summarise_all(mean)
#> Error in !funs: invalid argument type

# it's not a data.table issue
dtcars[, lapply(.SD, mean), by = cyl]
#>    cyl      mpg     disp        hp     drat       wt     qsec        vs
#> 1:   6 19.74286 183.3143 122.28571 3.585714 3.117143 17.97714 0.5714286
#> 2:   4 26.66364 105.1364  82.63636 4.070909 2.285727 19.13727 0.9090909
#> 3:   8 15.10000 353.1000 209.21429 3.229286 3.999214 16.77214 0.0000000
#>           am     gear     carb
#> 1: 0.4285714 3.857143 3.428571
#> 2: 0.7272727 4.090909 1.545455
#> 3: 0.1428571 3.285714 3.500000
```

To be clear, the same thing happens whether dtplyr is loaded or not, and given the `!` in the error looks like it's related to the rlang switchover. I'd post this as a dplyr issue, but since it's data.table-source specific, here seems appropriate.

Versions:
```r
sapply(c('dplyr', 'data.table', 'dtplyr'), packageVersion, simplify = FALSE)
#> $dplyr
#> [1] ‘0.5.0.9004’
[...]

Issue #17, opened 2016-03-08
mutate() loses groups


hadley:
Moved from https://github.com/hadley/dplyr/issues/919, by kykuo

``` R
library(dplyr)
library(dtplyr)

iris %>%
  group_by(Species) %>% 
  mutate(x = cumsum(Sepal.Length)) %>% 
  groups

data.table::data.table(iris) %>% 
  group_by(Species) %>%
  mutate(x = cumsum(Sepal.Length)) %>%
  groups
```

HuangNa2:
Seems like mutate reduces the group by 1 everytime its called
```
library(dplyr)
library(dtplyr)

iris %>%
  group_by(Species,Petal.Width,Sepal.Length) %>% 
  mutate(x = cumsum(Sepal.Length)) %>% 
  groups
```

Results in : 

[[1]]
Species

[[2]]
Petal.Width

From a bit of testing it appears that it removes the last grouped element each time mutate is called.

PR #55, opened 2017-09-13
WIP: from lazyeval to rlang

christophsax:
@krlmlr

- all verbs using rlang
- removed dependency on lazyeval
- added corresponding tests from dplyr: non working tests are commented out, should be removed and documented.
- should resolve several issues, e.g.,  #31, #54, #51, #43, #17.
- open / to be discussed: utility function `common_env()` does not always give the env we want. Using some variants in the code. Should be analyzed and unified.


lionel-:
Thanks for working on this Christoph.

Some stuff from my former attempt might be relevant: https://github.com/hadley/dtplyr/compare/master...lionel-:port-tidyeval?expand=1

In particular in the files `utils.R`, `utils-expr.R` and `predicates.R`. I don't remember much but I think instead of finding a common environment, I was flattening incoming quosures with `quo_expr(warn = TRUE)`.

hadley:
@christophsax it's been a while - are you interested in doing a little more to finish this off? I think it just needs a NEWS entry, and to make sure the tests that you've commented out are replaced with `skip()`s. Then I'll finish it off.

christophsax:
Yes, I can do that. Perhaps @krlmlr can also advise me on the details?

hadley:
Oh I looked more closely at the tests — it seems like you just copied and pasted the tests from dplyr, and you've already replaced the comments with skips.  In that case, all you need to do is add a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

It would also be useful if you could include `Fixes #123` for the issues that you can confirm this PR resolves.

christophsax:
Made a few adjustments to the tests, so the checks are passing again. I copy pasted and adjusted the tests for select again - something that probably should be done for the other verbs as well. Some tests need additional adjustments, a few are performed on tibbles or data.frames, so having them passing doesn't mean much.

I think some of the `group_by()` operations don't work as they should. 

Will look into both, but it will take me a few days.

hadley:
I think it makes sense to merge this now, and then work through verb-by-verb to reduce the tests to the minimum needed. Thanks for your work on this so far!

christophsax:
Sounds reasonable, and thanks for updating the news. I will do a separate PR for each verb I am working on.

PR #83, opened 2019-07-02
Closes #82 -- use nomatch=NULL for semi-joins

### LINKED ISSUE #82: nomatch = 0L is no longer idiomatic
As mentioned: https://github.com/Rdatatable/data.table/issues/3641

new idiom is using `nomatch = NULL`. Filing for a fix now.

MichaelChirico:
#82


hadley:
Thanks!

PR #87, opened 2019-07-23
Fixed "immutable" error message.

jonthegeek:
Might it be a good idea to switch the parameter to "mutable" to avoid the "not not" confusion that led to this?


hadley:
I thought about that but it feels weird to have `mutable = FALSE` as a default.

PR #118, opened 2019-11-03
Closes #112 -- if_else&coalesce use data.table equivalent

### LINKED ISSUE #112: Optimize if_else/coalesce to fifelse/fcoalesce?
`data.table` now has its own implementation of `if_else` (`fifelse`) and `coalesce` (`fcoalesce`).

I just tried the following: add two lines to `dplyr/tests/testthat.R`:

```
library(testthat)
library(dplyr)
if_else = data.table::fifelse
coalesce = data.table::fcoalesce

test_check("dplyr")
```

then run `testthat::test_dir('tests')`.

7 tests failed, but the same tests failed if I removed those two lines, so I'm taking that to mean `fifelse` and `fcoalesce` pass usage tests `dplyr` for behavior users expect of those functions.

Does it make sense/would it be straightforward to optimize those functions when used in `dtplyr`?

MichaelChirico:
Comments welcome. Was tempted to use `is_call(x, c('if_else', 'coalesce'))` but branching there to `quote(fifelse)` and `quote(fcoalesce)` would have been pretty ugly IINM.


R/tidyeval.R
```diff
@@ -66,6 +66,12 @@ dt_squash <- function(x, env, vars, j = TRUE) {
     } else if (is_call(x, "row_number", n = 1)) {
       arg <- dt_squash(x[[2]], vars = vars, env = env, j = j)
       expr(frank(!!arg, ties.method = "first", na.last = "keep"))
+    } else if (is_call(x, "if_else")) {
+      x[[1L]] = quote(fifelse)
```
hadley: `<-` 😉

MichaelChirico: done

hadley:
Thanks!

PR #121, opened 2019-11-10
Adds dt_pivot_wider() function that emulates tidyr::pivot_wider()

edgararuiz-zz:
- Attempts to follow the same implementation pattern for a new `step`.  This way it can be used in combination with other converted `dplyr` verbs:
  - Implements a new `step` via `step_pivot_wider()`, which contains the `new_step()` call
  - Implements a new `step_call()`,  with `dt_call.dtplyr_step_pivot_wider()`, which contains the actual `data.table` code that will be used
  - Implements `dt_pivot_wider()` which creates the step in for a lazy_dt()`
- Uses `data.table::dcast()` to perform the pivoting
- Documents function in order to avoid depending on `tidyr`'s help docs
- Names the function `dt_pivot_wider()`, and not `pivot_wider()`, to follow the same pattern as `dbplyr`: https://github.com/tidyverse/dbplyr/pull/344
- Emulates most of the tests `pivot_wider()` runs through in `tidyr`.  The new script has 100% coverage.
```
library(data.table)
library(tidyr)
library(dtplyr)
library(dplyr)

us_rent_income %>%
  as.data.table() %>%
  lazy_dt() %>%
  filter(as.integer(GEOID) < 5) %>%
  dt_pivot_wider(names_from = variable, values_from = c(estimate, moe), names_sep = "-") %>%
  filter(NAME == "Alaska")

#> Source: local data table [?? x 5]
#> Call:   dcast(data = `_DT1`[as.integer(GEOID) < 5], formula = GEOID + 
#>     NAME ~ variable, value.var = c(estimate = "estimate", moe = "moe"
#> ), fill = NULL, sep = "-")[NAME == "Alaska"]
#> 
#>   GEOID NAME   `estimate-income` `estimate-rent` `moe-income` `moe-rent`
#>   <chr> <chr>              <dbl>           <dbl>        <dbl>      <dbl>
#> 1 02    Alaska             32940            1200          508         13
#> 
```

A possible issue is that when there are only two columns in a data set, the pivoting adds a field called `.`.  When collected into `tibble` it creates a variable column, not sure if that's acceptable or not: 

```
tibble(key = c("x", "y", "z"), val = 1:3) %>%
  as.data.table() %>%
  lazy_dt() %>%
  dt_pivot_wider(names_from = key, values_from = val) %>%
  as_tibble()
#> # A tibble: 1 x 4
#>   .         x     y     z
#>   <chr> <int> <int> <int>
#> 1 .         1     2     3
```


edgararuiz:
ptal @hadley

hadley:
Now done in #189!

Issue #122, opened 2019-11-12
Using if_else() results in error 'could not find function "fifelse"'


larspijnappel:
I just installed `dtplyr` version 1.0.0 and it works like a charm. Unfortunately, when using the `dplyr::if_else()` function, it fails with the error message `could not find function "fifelse"`.

```r
library(dtplyr)
library(dplyr, warn.conflicts = FALSE)

mtcars %>% 
    lazy_dt() %>%
    mutate(gear2 = if_else(gear == 5, 10, gear)) %>%
    as_tibble()
```

hadley:
Try installing the latest version of data.table and see if works then?

larspijnappel:
Although my installed `data.table` package was already the latest version 1.12.6, I reinstalled it. 
I also upgraded `DT` from v0.9 to v10.0. 

But, after restarting, the issue remains..

hadley:
It definitely works for me:

``` r
library(dtplyr)
library(dplyr, warn.conflicts = FALSE)

mtcars %>% 
  lazy_dt() %>%
  select(gear) %>% 
  mutate(gear2 = if_else(gear == 4, 10, gear)) %>%
  as_tibble()
#> # A tibble: 32 x 2
#>     gear gear2
#>    <dbl> <dbl>
#>  1     4    10
#>  2     4    10
#>  3     4    10
#>  4     3     3
#>  5     3     3
#>  6     3     3
#>  7     3     3
#>  8     4    10
#>  9     4    10
#> 10     4    10
#> # … with 22 more rows
```

I'm not sure what's going on. Any ideas @MichaelChirico?

larspijnappel:
As pointed out by batpigandme in a related issue (https://github.com/tidyverse/dtplyr/issues/123#issuecomment-553114528) the solution is to add `library(data.table)`.

After re-reading your [blog](https://www.tidyverse.org/blog/2019/11/dtplyr-1-0-0/) I noticed I missed this advice:

> To use dtplyr, you must at least load dtplyr and dplyr. You might also want to load data.table so you can access the other goodies that it provides.

Sorry for overlooking this and thx for your help!

MichaelChirico:
would it make sense to detect data.table isn't loaded & include it in the lazy execution message?

hadley:
Or possibly inject the data table functions we use directly into the evaluation environment.

MichaelChirico:
[...]

PR #124, opened 2019-11-13
Add a check for data.table version before routing ifelse->fifelse

MichaelChirico:
Approach is to register the `data.table` version `onLoad` & keep it in a package-internal environment instead of fetching it each time in `dt_squash`.

Not sure if you have a different preferred approach for accomplishing this.

Related to #122


hadley:
Why not just bump the required version of data.table in the DESCRIPTION?

MichaelChirico:
That's also fine. Just offering the conservative approach :)

Feel free to close if you'd rather avoid the increase in complexity.

PR #125, opened 2019-11-13
typo in dt eval envir setname->setnames

### LINKED ISSUE #123: Using rename() results in error 'could not find function "setnames"'
Unfortunately,  `dtplyr` throws the following error when `rename()` is used: 

`Error in setnames(copy(`_DT1`), "mpg", "miles_per_gallon") : could not find function "setnames"`

```r
library(dtplyr)

mtcars %>% 
    lazy_dt() %>%
    rename(miles_per_gallon = mpg)
```

MichaelChirico:
Closes #123 

For tests, I guess this wasn't caught because `data.table` is loaded in `tests/testthat/helpers-library.R`.

Should we unload `data.table` for those tests? Or perhaps be more selective about loading `data.table` for tests?


hadley:
Perhaps we can evaluate the tests in an environment with parent above data.table in the search path? I'll think about this when I next work on dtplyr (probably at least a month away in order to give time for the issues to accumulate)

hadley:
It was easy enough to just not attach data.table for the tests. It didn't actually help here because I never actually executed the rename test, but should prevent similar problems in the future. Thanks for the fix!

Issue #85, opened 2019-07-03
Make it possible to switch between by and keyby


hadley:
[From](https://github.com/tidyverse/dtplyr/commit/7491a13ac48e27fc815eacfc246e83b534ef1cb8#r34166915) @jangorecki:

> Moreover doing keyby vs by adds an overhead, thus by should be preferred as default. Difference is not anything big (AFAIR up to 15-25%), and might depend on GForce being utilised. I understand that keyby would be consistent with how dplyr orders grouping results, but many (or even most?) of dplyr backends are not maintaining any specific order of results so that shouldn't be big deal.

PR #136, opened 2019-12-08
Support using by instead of keyby in grouped operations

psanker:
Addresses #85 

This amends how `keyby` is added to the translated calls by creating a separate structure for by/keyby and a linking function to link the object to the call. It's a bit more flexible than checking if there are groups and then adding them. Users are able to access `by=` by setting `key = FALSE` in `group_by()`:

```r
lazy_dt(mtcars, "mt") %>%
  group_by(cyl, key = FALSE) %>%
  summarize(mean_mpg = mean(mpg)) %>%
  show_query()

# mt[, .(mean_mpg = mean(mpg)), by = .(cyl)]
```
I'm still not certain about the term `key` for the parameter name since it could be confused with `key_by`. Maybe it could be something like `use_by` to be explicit?


NEWS.md
```diff
@@ -1,5 +1,8 @@
 # dtplyr (development version)
 
+*    `group_by()` now has a `key` parameter which, if set to `FALSE`, sets the
```
hadley: What about `reorder`? Or some thing that speaks directly to the action, rather than the underlying data.table implementation?

R/step-group.R
```diff
@@ -1,17 +1,54 @@
...
     parent,
     vars = parent$vars,
     groups = groups,
-    class = "dtplyr_step_group"
+    class = "dtplyr_step_group",
+    keyed = if (!is.null(keyed)) keyed else TRUE
   )
 }
 
+by_struct <- function(x) {
```
hadley: I think this would be a bit simpler if you made it a function that modifies a call, something like this:

```R
add_by <- function(x, groups, keyed) {
  if (length(groups) == 0) {
    return(x)
  }
  
  using <- if (keyed) "keyby" else "by"
  call[[using]] <- call2(".", !!!syms(groups))
  
  call
}
```

The name isn't quite right, but I think that approach is a bit simpler.

psanker: I agree! When I first started working on this, I thought the object under modification was a step, not the call itself, so I didn't want to introduce some weird side-effect. Since it's just a call, I removed the structure and replaced it with a function that's very similar to what you have written, which is in the latest push. I chose the name `add_grouping_parameter` for clarity.

NEWS.md
psanker: `reorder` works, but I feel like it could be misunderstood ("reorder from what?"). To follow the dplyr naming convention, I chose `arrange` instead. That should, in theory, be more understandable for people familiar with the tidyverse.

R/step-group.R
```diff
@@ -1,17 +1,30 @@
-step_group <- function(parent, groups = parent$groups) {
+step_group <- function(parent, groups = parent$groups, keyby = parent$keyby) {
```
hadley: Would you mind also renaming this `keyby` argument to arrange? I think it will make the code a bit easier to read.

vignettes/translation.Rmd
```diff
@@ -171,7 +171,13 @@ Just like in dplyr, `group_by()` doesn't do anything by itself, but instead modi
 dt %>% group_by(a) %>% summarise(b = mean(b)) %>% show_query()
 ```
 
-(Currently there's no way to use `by` instead of `keyby`, but that is planned [for the future](https://github.com/tidyverse/dtplyr/issues/85).)
+You may use `by` instead of `keyby` if you set `key = FALSE`:
```
hadley: Needs an update?

R/step-group.R
```diff
@@ -1,17 +1,30 @@
...
     parent,
     vars = parent$vars,
     groups = groups,
-    class = "dtplyr_step_group"
+    class = "dtplyr_step_group",
+    keyby = keyby
   )
 }
 
+add_grouping_parameter <- function(call, groups, keyby) {
```
hadley: Maybe just `add_grouping_param()`? And I notice that all the calls look like `add_grouping_parameter(something, x$groups, x$keyby)` so maybe the second argument should just be a step object?

psanker: Done and done!

vignettes/translation.Rmd
psanker: Yes it does. Fixed in latest commit.

R/step-group.R
psanker: Done!

hadley:
Thanks!

GitHunter0:
@psanker , thank you, this was a good issue to solve. 

The only problem left is that `dplyr::group_by()` does not have the parameter `key`, which makes `dtplyr` and `dplyr` code not perfectly interchangeable...

PR #162, opened 2020-08-08
Add step_relocate

smingerson:
Combines code from `step_select()` and `dplyr::relocate()` to create `step_relocate()`. I added one basic test, unsure if more are needed. Also thought that maybe `@import dplyr relocate` needs to happen conditioned on dplyr version, but not quite sure how to handle that.

Edit: The failing tests are due to a change in tibble 3.0:  glimpse() uses "Rows" and "Columns" instead of "Variables" and "Observations", because we're not sure if the data is tidy here (#614).


R/step-subset.R
```diff
@@ -135,6 +135,47 @@ select.dtplyr_step <- function(.data, ...) {
   step_group(out, groups)
 }
 
+#' @importFrom dplyr relocate
```
hadley, line 138: I think we should probably just bump the dplyr dependency in the DESCRIPTION to 1.0.0

R/step-subset.R
```diff
@@ -135,6 +135,47 @@ select.dtplyr_step <- function(.data, ...) {
...
+  if (length(to_move) == 0) {
+    return(.data)
+  }
+  .before <- enquo(.before)
+  .after <- enquo(.after)
+  has_before <- !quo_is_null(.before)
+  has_after <- !quo_is_null(.after)
+  if (has_before && has_after) {
+    abort("Must supply only one of `.before` and `.after`.")
+  }
```
hadley, line 153: Can you please follow https://style.tidyverse.org and put the `else` on the same line as the `}`?

R/step-subset.R
```diff
@@ -135,6 +135,47 @@ select.dtplyr_step <- function(.data, ...) {
...
+#' @importFrom dplyr relocate
+#' @export
+relocate.dtplyr_step <- function(.data, ..., .before = NULL, .after = NULL) {
+  vars <- set_names(.data$vars, .data$vars)
+  to_move <- unname(tidyselect::eval_select(expr(c(...)), data = vars))
+  groups <- .data$groups
+  if (length(to_move) == 0) {
+    return(.data)
+  }
+  .before <- enquo(.before)
```
hadley, line 147: What about just doing `.before <- unname(tidyselect::eval_select(.before, vars))`? I think that would simplify the code quite a bit.

tests/testthat/test-step-subset.R
```diff
@@ -38,6 +38,11 @@ test_that("simple calls generate expected translations", {
     expr(DT[, .(a = x, y)])
   )
 
+  expect_equal(
```
hadley, line 41: I think `relocate()` is complicated enough that it should have it's own test case — it needs to have an example of exactly one of `.before` and `.after`, neither of them, and both of them. I'd also look at https://github.com/tidyverse/dplyr/blob/master/tests/testthat/test-relocate.R to see if there are other tricky cases we should also test for here.

smingerson: I've added tests based off dplyr, as suggested. I also added one to ensure relocation does not change the grouping variables.

R/step-subset.R
smingerson: Could you briefly elaborate what you expect to be simplified away? The code is only very lightly modified from dplyr's relocate, so I had been assuming it was structured as such for good reasons.

hadley: Oh well if it matches dplyr, keep it like that. I have no recollection of what the dplyr code looks like 😆

PR #189, opened 2021-02-13
Implement pivot_wider()

markfairbanks:
Implements `pivot_wider()` for use on lazy data.tables.

A few notes:
* For `names_repair` to be an option, `.name_repair` needed to be exposed in `as_tibble.dtplyr_step()`. It doesn't break any tests, but `print.dtplyr_step()` and `dtplyr:::simulate_vars()` no longer do checks for valid names.
* I used `lapply()` & `vapply()` - not sure if you'd rather use [compat-purrr.R](https://github.com/r-lib/rlang/blob/master/R/compat-purrr.R) instead.
* It currently doesn't work with `".value"` in `names_from`. I'm not positive what the logic is for that.
* All tests are taken from the tidyr version. It works in all cases except for 3 that are commented out:
  - One with `".value"` usage
  - One that deals with data frame columns
  - One that returns a list column
* In the case of multiple `names_from` columns or in the case that `names_glue` is used, deep copies are made when using `mutate()` to help extract the new column names. This could be made faster by using [data.table's shallow copy trick.](https://github.com/Rdatatable/data.table/issues/3214#issuecomment-462490046)
According to Matt Dowle it is supposed to be supported until they export `shallow()`. Not sure your opinion on that. `pivot_wider()` works as is, or a simple internal `shallow_dt <- function(x) filter(x, TRUE)` could be used here.

If you have any changes you want me to make let me know. Also feel free to make changes yourself if you want.


R/step-call-pivot_wider.R
```diff
@@ -0,0 +1,220 @@
...
+                                   names_prefix = "",
+                                   names_sep = "_",
+                                   names_glue = NULL,
+                                   names_sort = FALSE,
+                                   names_repair = "check_unique",
+                                   values_from = value,
+                                   values_fill = NULL,
+                                   values_fn = NULL,
+                                   ...) {
+  data <- lazy_dt(data)
```
hadley, line 188: Can you add one test for this, just to make sure it's correctly plumbed together? (it's easy to make mistakes in the tidy eval forwarding)

tests/testthat/test-step-call-pivot_wider.R
```diff
@@ -0,0 +1,240 @@
+test_that("can pivot all cols to wide", {
```
hadley, line 1: It would be useful to add a collection of snapshot tests that capture the generated expressions for key tasks. That makes it possible to review the translations, not just the results.

tests/testthat/test-step-call-pivot_wider.R
```diff
@@ -0,0 +1,240 @@
...
+# https://github.com/tidyverse/tidyr/issues/804
+test_that("column with `...j` name can be used as `names_from`", {
+  df <- lazy_dt(tibble(...8 = c("x", "y", "z"), val = 1:3))
+  step <- pivot_wider(df, names_from = ...8, values_from = val)
+  pv <- collect(step)
+  expect_equal(step$vars, c("x", "y", "z"))
+  expect_equal(nrow(pv), 1)
+})
+
+# test_that("data frame columns pivot correctly", {
```
hadley, line 58: I think you can just delete this

tests/testthat/test-step-call-pivot_wider.R
```diff
@@ -0,0 +1,240 @@
...
+#
+#   out <- pivot_wider(df, names_from = g, values_from = d)
+#   expect_equal(out$a$x, 1:2)
+#   expect_equal(out$b$y, 7:8)
+# })
+
+
+# column names -------------------------------------------------------------
+
+# test_that("names_glue affects output names", {
```
hadley, line 73: I'd just delete this since you've replaced it below

tests/testthat/test-step-call-pivot_wider.R
```diff
@@ -0,0 +1,240 @@
...
+  pv <- df %>%
+    pivot_wider(id_cols = name, names_from = var, values_from = value) %>%
+    collect()
+  expect_equal(nrow(pv), 2)
+})
+
+
+# non-unique keys ---------------------------------------------------------
+
+# test_that("duplicated keys produce list column with warning", {
```
hadley, line 133: I'd delete this, and then mention the difference in behaviour in the docs.

hadley (commented):
Looks like a great start — thanks for working on this!

It would be great if we could get `.value` to work, because I'm pretty sure this feature was originally inspired by data.table. I can try re-remember what this does if it you need some more hints.

I think it's ok to use the shallow copy trick, but it would be good to wrap it up in a dtplyr function, so that when data.table exports shallow we only need to change one place.

DESCRIPTION
```diff
@@ -26,7 +26,9 @@ Imports:
     lifecycle,
     rlang,
     tibble,
-    tidyselect
+    tidyr (>= 1.1.0),
```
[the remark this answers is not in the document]
markfairbanks: > If not, I'm happy to explain.

Yep, that'd be great. Didn't know this was possible.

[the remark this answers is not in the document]
hadley: The basic idea is implemented in `vctrs::s3_register()` — you copy the source of that file into the package, and then call `s3_register()` in `.onLoad()`. That automatically sets up the hooks to register dtplyr's methods when tidyr is attached.

markfairbanks:
Alright @hadley - all set for your review! Pretty sure I covered everything, but let me know if there's anything else.

R/step-call-pivot_wider.R
```diff
@@ -0,0 +1,225 @@
...
+    new_vars <- unlist(new_vars)
+  }
+
+  no_id <- length(id_cols) == 0
+
+  if (no_id) {
+    lhs <- "..."
+    new_vars <- c(".", new_vars)
+  } else {
+    lhs <- paste(id_cols, collapse = " + ")
```
hadley, line 97: Is there a reason to make these strings rather than calls? The main downside of using strings is that there's a high likelihood of weirdness in the presence of non-syntactic names.

R/step-call-pivot_wider.R
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  if (length(values_from) > 1) {
+    new_vars <- lapply(values_from, function(.x) paste(.x, new_vars, sep = names_sep))
+    new_vars <- unlist(new_vars)
+  }
+
+  no_id <- length(id_cols) == 0
+
+  if (no_id) {
```
hadley, line 93: I'd turn this into one paragraph too

R/step-call-pivot_wider.R
```diff
@@ -0,0 +1,225 @@
...
+
+  dcast_form <- paste(lhs, rhs, sep = " ~ ")
+
+  args <- list(
+    formula = dcast_form,
+    value.var = values_from,
+    fun.aggregate = values_fn,
+    sep = names_sep,
+    fill = values_fill
+  )
```
hadley, line 112: And this. If you don't use `dcast_form` elsewhere, I'd be tempted to inline it into the `list()` call

R/step-call-pivot_wider.R
```diff
@@ -0,0 +1,225 @@
...
+      args = list(old = new_vars, new = glue_vars),
+      vars = c(id_cols, glue_vars),
+      in_place = FALSE
+    )
+
+    # In case of names_sort = TRUE
+    new_vars <- glue_vars
+  }
+
+  if (nchar(names_prefix) > 0 && is.null(names_glue)) {
```
hadley, line 142: I think the structure here should be:

```R
if (!is.null(names_glue)) {

} else if (nchar(names_prefix) > 0) {

}
```

R/zzz.R
```diff
@@ -9,6 +9,10 @@
   register_s3_method("dplyr", "intersect", "dtplyr_step")
   register_s3_method("dplyr", "setdiff", "dtplyr_step")
   register_s3_method("dplyr", "union", "dtplyr_step")
+
+  register_s3_method("tidyr", "pivot_wider", "data.table")
```
hadley, line 13: I think I'd add these at beneath the corresponding dplyr methods.

tests/testthat/test-step-call-pivot_wider.R
```diff
@@ -0,0 +1,210 @@
...
+
+test_that("names_glue affects output names & auto-converts data.table to lazy_dt", {
+  df <- data.table(
+    x = c("X", "Y"),
+    y = 1:2,
+    a = 1:2,
+    b = 1:2
+  )
+
+  step <- pivot_wider(df, names_from = x:y, values_from = a:b, names_glue = "{x}{y}_{.value}")
```
hadley, line 76: Here I'd suggest an `expect_snapshot(show_query(step))`, just to capture the sequence of calls that you're generating.  Similarly for the test for `names_sort = TRUE`

hadley (commented):
Looking great — just a few more smaller comments.

R/step-call-pivot_wider.R
markfairbanks: I think the change I made here covers what you were looking for, let me know if I misunderstood.

In general all changes are implemented and ready for review.

tests/testthat/_snaps/step-call-pivot_wider.md
```diff
@@ -0,0 +1,17 @@
...
+      "b"))[, .(a_X_1, a_Y_2, b_X_1, b_Y_2)], old = c("a_X_1", "a_Y_2", 
+      "b_X_1", "b_Y_2"), new = c("X1_a", "Y2_a", "X1_b", "Y2_b"))
+
+# can sort column names
+
+    Code
+      show_query(step)
+    Output
+      setcolorder(dcast(DT, formula = "... ~ chr", value.var = "int")[, 
+          .(Wed, Tue, Mon)], c("Mon", "Tue", "Wed"))
```
hadley, line 16: Could we not generate:

```R
cast(DT, formula = "... ~ chr", value.var = "int")[,  (Mon, Tue, Wed)]
```

Or does that make the renaming code too complicated?

R/step-call-pivot_wider.R
```diff
@@ -0,0 +1,220 @@
...
+    new_vars <- lapply(values_from, function(.x) paste(.x, new_vars, sep = names_sep))
+    new_vars <- unlist(new_vars)
+  }
+
+  no_id <- length(id_cols) == 0
+  if (no_id) {
+    lhs <- "..."
+    new_vars <- c(".", new_vars)
+  } else {
+    lhs <- paste(syms(id_cols), collapse = " + ")
```
hadley, line 96: I think this will still make `lhs` a string. I'd use `Reduce` + `call2` instead:

``` r
library(rlang)
id_cols <- letters[1:3]
Reduce(function(x, y) call2("+", x, y), syms(id_cols))
#> a + b + c
```

<sup>Created on 2021-02-16 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

It's possible I already have a helper somewhere for this 

...

Hmmm, doesn't look like it, although I use the same pattern in `dt_squash_if()` so it might be worth extracting out something like this:

```R
call_reduce <- function(x, fun) {
  Reduce(function(x, y) call2(fun, x, y), x)
}
```

R/step-call-pivot_wider.R
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  vars <- c(id_cols, new_vars)
+
+  rhs <- paste(syms(names_from), collapse = " + ")
+
+  args <- list(
+    formula = paste(lhs, rhs, sep = " ~ "),
```
hadley, line 104: And this would be `call2("~", lhs, rhs)`

tests/testthat/_snaps/step-call-pivot_wider.md
markfairbanks: All set.

FYI - there are two cases if `names_sort = TRUE`.

Case 1: `length(id_cols) == 0` (like the snapshot above)
The translation now reads like you recommended (only uses the "select" step, no `setcolorder()` call).

Case 2: `length(id_cols) > 0`
There is no "select" step necessary in this case, so it still uses `setcolorder()` for names sorting.

If you'd rather have Case 2 use "select" as well let me know.

R/step-call-pivot_wider.R
markfairbanks: All set on the formula changes too. I ended up having to leave `lhs` as a string only in the case where there were no id cols and `lhs <- "..."`. data.table was throwing this error otherwise:

> Error in check_formula(formula, names(data), valnames) : '...' used in an incorrect context

hadley:
Thanks for all your work on this! I finished it off with a couple of unit tests to ensure that we had full code coverage, and to make sure that I had at least a basic understanding of what was going on.

Hopefully this PR didn't feel like too much of a slog, but I think it's worth being especially thorough on your first PR in the hopes that it will make future PRs easier and faster 😄

markfairbanks:
Nope, it wasn't bad at all! Thanks for taking the time to do the reviews. As you mentioned, I think future PRs should now be a bit faster.

PR #191, opened 2021-02-17
Enable case_when(TRUE ~ default) syntax

markfairbanks:
Currently data.table [doesn't recycle the left hand side "conditions" in `fcase()`](https://github.com/Rdatatable/data.table/issues/4258). This causes `case_when(x == 1 ~ 1, TRUE ~ 2)` syntax to fail in the current translation. `fcase()` also doesn't currently allow a vector for the `default` arg.

This implements a simple workaround that is mentioned in that issue.

### DIFF
--- a/R/tidyeval.R
+++ b/R/tidyeval.R
@@ -110,14 +110,12 @@ dt_squash_call <- function(x, env, data, j = TRUE) {
   } else if (is_call(x, "case_when")) {
     # case_when(x ~ y) -> fcase(x, y)
     args <- unlist(lapply(x[-1], function(x) {
-      list(x[[2]], x[[3]])
+      list(
+        if (isTRUE(x[[2]])) call2("rep", TRUE, quote(.N)) else x[[2]],
+        x[[3]]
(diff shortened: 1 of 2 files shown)


R/tidyeval.R
```diff
@@ -113,6 +113,11 @@ dt_squash_call <- function(x, env, data, j = TRUE) {
       list(x[[2]], x[[3]])
     }))
     args <- lapply(args, dt_squash, env = env, data = data, j = j)
+    odd_index <- as.logical(seq_along(args) %% 2)
+    args[odd_index] <- lapply(
+      args[odd_index],
+      function(.x) if (isTRUE(.x)) call2("rep", TRUE, quote(.N)) else .x
```
hadley, line 119: I think this would be cleaner to apply to `x[[2]]` above — then you wouldn't need to compute the odds index.

markfairbanks: Yep, makes sense. All set.

hadley:
Thanks!

PR #197, opened 2021-02-22
Implement fill()

markfairbanks:


tests/testthat/_snaps/step-call-pivot_wider.md
```diff
@@ -3,7 +3,7 @@
     Code
       show_query(step)
     Output
-      setnames(dcast(`_DT5`, formula = "..." ~ x + y, value.var = c("a", 
+      setnames(dcast(`_DT6`, formula = "..." ~ x + y, value.var = c("a",
```
hadley, line 6: Can you please explicitly set the name in the `lazy_dt()` call to stop this from changing in the future?

tests/testthat/test-fill.R
```diff
@@ -0,0 +1,72 @@
...
+    dbl = c(NA_real_, NA),
+  )
+  dt <- lazy_dt(tbl, "DT")
+
+  down <- fill(dt, int, dbl)
+  up <- fill(dt, int, dbl, .direction = "up")
+
+  expect_identical(collect(down), tbl)
+  expect_identical(collect(up), tbl)
+})
```
hadley, line 13: I think you can remove this, since it's mostly testing the behaviour of `nafill()`?

tests/testthat/test-fill.R
```diff
@@ -0,0 +1,72 @@
...
+  expect_equal(out$int, c(1L, 1L))
+  expect_equal(out$dbl, c(1, 1))
+})
+
+test_that("fill respects grouping", {
+  tbl <- tibble(x = c(1, 1, 2), y = c(1, NA, NA))
+  dt <- lazy_dt(tbl, "DT")
+  out <- dt %>% dplyr::group_by(x) %>% fill(y) %>% collect()
+  expect_equal(out$y, c(1, 1, NA))
+})
```
hadley, line 72: Same here; I think these are redundant.

tests/testthat/_snaps/step-call-pivot_wider.md
markfairbanks: This is the snapshot that is generated when testing that `pivot_wider.data.table()` redirects to `pivot_wider.dtplyr_step()`. Do you have a preferred name for `lazy_dt()` to use inside of `pivot_wider.data.table()` and future `.data.table` methods?

hadley: Oooooooh. Leave it as for now then, and at some point in the future we should figure out a better solution.

markfairbanks: Sounds good - all set for review

PR #199, opened 2021-02-23
Don't include vars deleted in mutate() in step$vars

markfairbanks:
reprex of issue fixed
``` r
library(dtplyr)
library(dplyr)

dt <- lazy_dt(tibble(x = 1:3, y = 1:3))

step <- dt %>%
  mutate(x = NULL)

# Correct translation/output
step
#> Source: local data table [3 x 1]
#> Call:   copy(`_DT1`)[, `:=`(x = NULL)]
#> 
#>       y
#>   <int>
#> 1     1
#> 2     2
#> 3     3
#> 

# Still includes dropped vars
step$vars
#> [1] "x" "y"
```

<sup>Created on 2021-02-23 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>


hadley:
Good catch, thanks!

PR #204, opened 2021-03-01
Implement pivot_longer()

markfairbanks:
A few notes:
* I added `vec_cast()` to `add_dt_wrappers()` - I wasn't sure if there was a different way you'd rather add that.
* I implemented a version of `pmap_chr()` taken from [rlang's `compat-purrr.R`](https://github.com/r-lib/rlang/blob/master/R/compat-purrr.R) file for use in `str_extract()` to avoid a purrr dependency.


R/step-call-pivot_longer.R
```diff
@@ -0,0 +1,388 @@
...
+                                     names_transform = list(),
+                                     names_repair = "check_unique",
+                                     values_to = "value",
+                                     values_drop_na = FALSE,
+                                     values_ptypes = list(),
+                                     values_transform = list(),
+                                     ...) {
+
+  sim_data <- simulate_vars(data)
+  sim_vars <- names(sim_data)
```
hadley, line 63: I'd move this definition closer to where it's used.

R/step-call-pivot_longer.R
```diff
@@ -0,0 +1,388 @@
...
+    na.rm = values_drop_na,
+    variable.factor = FALSE
+  )
+
+  # Clean up call args if defaults are used
+  if (variable_name == "variable") {
+    args$variable.name <- NULL
+  }
+
+  if (length(values_to) == 1) {
```
hadley, line 116: `identical(values_to, "value")` is a slightly simpler way of expressing this

R/step-call-pivot_longer.R
```diff
@@ -0,0 +1,388 @@
...
+    out <- mutate(out, !!!into_cols)
+
+    # Need to drop variable_name and move names_to vars to correct position
+    # Recreates relocate logic so only select is necessary, not relocate + select
+    out_vars <- out$vars
+    var_idx <- which(out_vars == variable_name)
+    before_vars <- out_vars[seq_along(out_vars) < var_idx]
+    after_vars <- out_vars[seq_along(out_vars) > var_idx]
+
+    out <- select(out, !!!syms(before_vars), !!!syms(names_to), !!!syms(after_vars))
```
hadley, line 154: Can this use `setcolorder()`?

R/step-call-pivot_longer.R
```diff
@@ -0,0 +1,388 @@
...
+  } else if (!multiple_names_to && uses_dot_value) {
+    out <- mutate(out, variable = NULL)
+  }
+
+  out <- step_repair(out, repair = names_repair)
+
+  ## names_ptype & names_transform
+  cast_vars <- intersect(names_to, names(names_ptypes))
```
hadley, line 162: This feels sort of fundamentally un-data.table-y to me. Do you think maybe it's better to just not support these arguments?

tests/testthat/_snaps/step-call-pivot_longer.md
```diff
@@ -0,0 +1,17 @@
+# can coerce values cols
+
+    Code
+      show_query(step)
+    Output
+      melt(DT, measure.vars = c("x", "y"), variable.name = "name", 
+          variable.factor = FALSE)[, `:=`(value = .Primitive("as.character")(value))]
```
hadley, line 7: Looks like `as.character()` got inlined here?

hadley (commented):
Looks great, thanks! A few minor comments/questions below.

R/step-call-pivot_longer.R
markfairbanks: There are two things this part does: 1 - drops the `variable_name` column, and 2 - puts the new `names_to` columns where the `variable_name` column used to be. `setcolorder()` doesn't allow you to drop columns, only rearrange them. 

It's built out this way so that it only uses `select()`, not `select()` + `setcolorder()`. Do you want to leave it as is, or would you rather use both functions?

R/step-call-pivot_longer.R
markfairbanks: I think this one is up to your preference. I think you're right - this is functionality that starts to split from what would be considered "normal" data.table. But if you want users of dtplyr to be able to quickly add `lazy_dt()`/`as_tibble()` to their currently existing pipe chains it could be worth keeping.

What do you prefer? Remove or keep them? And if "remove", do you want me to add error messages for them?

mgirlich:
In hindsight of making `tidyr::pivot_longer_spec()` and `tidyr::build_longer_spec()` generics (see https://github.com/tidyverse/tidyr/pull/1101) it would probably make sense to internally split this function into `dtplyr_pivot_longer_spec()` and `dtplyr_build_longer_spec()`. This makes it easier to convert them to the corresponding generics as soon as they are converted to S3 methods. The same holds true for `pivot_wider()`

R/step-call-pivot_longer.R
hadley: Ah ok, makes sense. We can leave as is.

R/step-call-pivot_longer.R
hadley: I think remove — they're not particularly useful arguments so I think few people will use them. You can make an error like "`values_ptype` not supported by dtplyr". 

In general, before embarking on any particularly complex translations (or translations that don't have a good data.table equivalent), I think it's reasonable to determine if people are actually using them. There's little point solving a problem that affects few people when there are missing translations for popular tidyr components.

markfairbanks: Sounds good. And just to make sure before I change anything - am I dropping both `_ptype` and the `_transform` arguments, or just the `_ptype`?

hadley: Definitely `_ptype`. I think you could probably skip `_transform` too — I think doing it properly probably means using NSE so you can inline the transformations in to the steps.

hadley: (Like in `across()` and somewhere else I can't remember)

hadley:
When I document this locally, I don't get the `\method{}` syntax in the .Rd files? Are you doing something special to make this work?

markfairbanks:
I was actually running into the same thing when I just tried, but it looks like you just have to run `library(tidyr)`. I must have had tidyr loaded in the session to do some tests when I was documenting them originally.

hadley:
Ah of course; and that never happens for me because I always run in a clean session.

hadley:
Used a hack from @gaborcsardi  to always attach tidyr when documenting.

hadley:
Thanks!

PR #207, opened 2021-03-02
`lazy_dt()` keeps groups

### LINKED ISSUE #206: `lazy_dt()` doesn't respect grouping
The grouping information is lost after using `lazy_dt()`

``` r

tibble(x = 1) %>% 
  group_by(x) %>% 
  lazy_dt()
#> Call:   `_DT1`
#> 
#>       x
#>   <dbl>
#> 1     1
#> 
```

<sup>Created on 2021-03-02 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

mgirlich:
Fixes #206


R/step-first.R
```diff
@@ -61,6 +61,12 @@
 #'   filter(mpg < mean(mpg)) %>%
 #'   summarise(hp = mean(hp))
 lazy_dt <- function(x, name = NULL, immutable = TRUE, key_by = NULL) {
+  if (is.data.frame(x)) {
```
hadley, line 64: You don't need this if statement since `group_vars()` already returns `character()` when applied to a data frame.

mgirlich: ah, I should have added a comment there. I added this line because of a test in `test-step_call.R` where `lazy_dt()` is applied to a matrix. As there is no method `group_vars.matrix()` this errored. Do you prefer to add a method for matrices instead?

hadley: If we added a method it would have to go into dplyr, which would be a pain.

Maybe just flip it and make the test `if (inherits(x, "grouped_df"))` ?

hadley: Or `groups <- tryCatch(group_vars(x), error = function(e) character())`? 

I don't generally like `tryCatch()` for control flow, but I think it makes sense here since there's no requirement for `as.data.table()` and `group_vars()` to provide methods for the same classes.

mgirlich: Makes sense to me. I also added a comment for clarification.

R/step-first.R
```diff
@@ -61,6 +61,9 @@
 #'   filter(mpg < mean(mpg)) %>%
 #'   summarise(hp = mean(hp))
 lazy_dt <- function(x, name = NULL, immutable = TRUE, key_by = NULL) {
+  # use `tryCatch()` in case there is no matching `group_vars()` method
```
hadley, line 64:
```suggestion
  # in case `x` has `as.data.table()` method but not a `group_vars()` method 
```

PR #209, opened 2021-03-02
Fix join column order

### LINKED ISSUE #139: Left join columns order changed
Thank you for all the work put into dtplyr; it already is saving me loads of time in my day to day. 

It appears that left_join causes columns to change order. Doesn't appear to be an issue with right_join or merge.

``` r
library(data.table)
library(dtplyr)
library(dplyr,warn.conflicts=FALSE)

packageVersion("data.table")
#> [1] '1.12.8'
packageVersion("dtplyr")
#> [1] '1.0.0.9000'

x1 <- data.frame(x = 1:3, y = 1)
x2 <- data.frame(x = 1:3, z = 2)

x1 %>%
  left_join(x2, by="x")
#>   x y z
#> 1 1 1 2
#> 2 2 1 2
#> 3 3 1 2

x1 %>%
  lazy_dt() %>%
  left_join(x2, by="x") %>%
  as_tibble()
#> # A tibble: 3 x 3
#>       x     z     y
#>   <int> <dbl> <dbl>
#> 1     1     2     1
#> 2     2     2     1
#> 3     3     2     1
```

<sup>Created on 2019-12-29 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

mgirlich:
Fixes #139 .

The idea is to simply 

* make sure `tbl_vars()` gives the variables in the same order as for `dplyr` in `step_subset_on()` and `step_join()`.
* add a select statement to `print()`, `show_query()`, and `collect()`.

I am not entirely sure if this works, as I cannot really run the tests right now. The issue is that this currently always select whether needed or not. This implies that quite a lot of tests need to be adapted and snapshots need to be checked.

Also this is basically a breaking change but I think the consistency between `dplyr` and `dtplyr` is still preferable.

a) do you see a simpler way of doing this?
b) do you mind always adding a select statement whether needed or not?


hadley:
Why can't you run the tests?

markfairbanks:
Instead of requiring an extra select statement on all queries, would it be easier to just add a call to `setcolorder()` in the case of a "simple" left join?

```r
left_join.dtplyr_step <- function(x, y, ..., by = NULL, copy = FALSE, suffix = c(".x", ".y")) {
  y <- dtplyr_auto_copy(x, y, copy = copy)
  by <- dtplyr_common_by(by, x, y)

  if (join_is_simple(x$vars, y$vars, by)) {
    col_order <- unique(c(x$vars, y$vars))
    out <- step_subset_on(y, x, i = y, on = by)

    step_call(
      out,
      "setcolorder",
      args = list(col_order),
      vars = col_order,
      in_place = FALSE
    )
  } else {
    step_join(x, y, on = by, style = "left", suffix = suffix)
  }
}
```

That way it avoids another deep copy of the data.table and other queries can remain the same.

mgirlich:
true, that's much simpler! I somehow thought this also affects the `step_join()` path but it actually doesn't. Nice one!

mgirlich:
@markfairbanks As the idea comes fully from you I would be happy to see your name instead of mine for this fix in the `NEWS.md`. If you are fine with that I'll add the corresponding news entry.

markfairbanks:
I'm good with that @mgirlich, thanks :thumbsup:

R/step-join.R
```diff
@@ -88,7 +88,16 @@ left_join.dtplyr_step <- function(x, y, ..., by = NULL, copy = FALSE, suffix = c
   by <- dtplyr_common_by(by, x, y)
 
   if (join_is_simple(x$vars, y$vars, by)) {
-    step_subset_on(y, x, i = y, on = by)
+    col_order <- unique(c(x$vars, y$vars))
+    out <- step_subset_on(y, x, i = y, on = by)
+
+    step_call(
```
hadley, line 94: I think it's probably time to make `step_col_order()`. Would you mind doing that and using it here and in `pivot_wider()`? 

@markfairbanks should we also use it in `relocate()`?

tests/testthat/test-step-join.R
```diff
@@ -80,7 +80,7 @@ test_that("simple left joins use [", {
 
   expect_equal(
     dt1 %>% left_join(dt2, by = "x") %>% show_query(),
-    expr(dt2[dt1, on = .(x), allow.cartesian = TRUE])
+    call2("setcolorder", expr(dt2[dt1, on = .(x), allow.cartesian = TRUE]), c("x", "a", "b"))
```
hadley, line 83: Why are you using `call2()` here?

mgirlich: I first had the simpler form without `call2()` but this then the argument `c("x", "a", "b")` differs (language vs character vector)

``` r
library(rlang)
e <- expr(setcolorder(dt2[dt1, on = .(x), allow.cartesian = TRUE], c("x", "a", "b")))
e2 <- call2("setcolorder", expr(dt2[dt1, on = .(x), allow.cartesian = TRUE]), c("x", "a", "b"))

str(e[[3]])
#>  language c("x", "a", "b")
str(e2[[3]])
#>  chr [1:3] "x" "a" "b"
```

<sup>Created on 2021-03-03 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

hadley:
```R
e <- expr(setcolorder(dt2[dt1, on = .(x), allow.cartesian = TRUE], !!c("x", "a", "b")))
```

😉

mgirlich: I thought it was problematic to use `!!` in `expect_equal()` but actually it was only in `expect_snapshot()` and this works nicely!

R/step-join.R
mgirlich: @hadley done. This was actually easier than expected 😄

tests/testthat/test-step-join.R
hadley: Oh yeah I need to fix that 😞

R/step-join.R
markfairbanks: @hadley Yep, I think we should. It appears to be decently faster than using "select". I'll work on a PR.

``` r
library(data.table)

data_size <- 100000
test_df <- 
  data.table(
    x = sample(1:5, data_size, TRUE),
    y = sample(letters[1:5], data_size, TRUE),
    z = sample(LETTERS[1:5], data_size, TRUE)
  )

bench::mark(
  select = test_df[, c("z", "y", "x")],
  setcolorder = setcolorder(copy(test_df), c("z", "y", "x"))[]
)
#> # A tibble: 2 x 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 select        1.35ms   1.59ms      561.    3.45MB     25.5
#> 2 setcolorder 276.24µs 400.54µs     2024.    1.94MB    119.
```

hadley: Cool, thanks!

PR #212, opened 2021-03-03
Filtering by negated logical column

### LINKED ISSUE #211: Filtering by negated logical column
While filtering by a logical column is nicely wrapped, it doesn't work when negating a logical column:

``` r
library(dtplyr)

data.table::data.table(lgl_col = c(TRUE, FALSE)) %>% 
  filter(!lgl_col)
#> Error in `[.data.table`(`_DT1`, !lgl_col): lgl_col is not found in calling scope but it is a column of type logical. If you wish to select rows where that column contains TRUE, or perhaps that column contains row numbers of itself to select, try DT[(col)], DT[DT$col], or DT[col==TRUE] is particularly clear and is optimized. When the first argument inside DT[...] is a single symbol (e.g. DT[var]), data.table looks for var in calling scope.
```

<sup>Created on 2021-03-03 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

mgirlich:
fixes #211


R/step-subset-filter.R
```diff
@@ -32,6 +32,24 @@ filter.dtplyr_step <- function(.data, ..., .preserve = FALSE) {
...
+  }
+
+  dot <- dots[[1]]
+
+  if (is_symbol(dot)) {
+    return(TRUE)
+  }
+
+  # catch expressions of form `!x`
+  is_call(dot) &&
```
hadley, line 47: You can simplify this a bit by using more arguments to `is_call()`

mgirlich: neat, actually this makes so much sense that I could have expected that from an `rlang` function!

R/step-subset-filter.R
```diff
@@ -32,6 +32,21 @@ filter.dtplyr_step <- function(.data, ..., .preserve = FALSE) {
...
+filter_by_lgl_col <- function(dots) {
+  if (length(dots) > 1) {
+    return(FALSE)
+  }
+
+  dot <- dots[[1]]
+
+  if (is_symbol(dot)) {
+    return(TRUE)
+  }
```
hadley, line 44: This is a super small comment, but I've been thinking about this lately — I'd suggest more clearly making this a "paragraph" of code by removing the blank line on line 41. I think that it makes these two statements hang together better as a related thought.

mgirlich: Thinking about it I agree 👍

PR #219, opened 2021-03-05
slice_head() speed up

mgirlich:
it also makes sense to speed up `slice_head()` and `slice_tail()` by using `.I`


hadley (commented):
Can you tweak the docs now too please? I think now all the slice function use `i`?

mgirlich:
Yes, they do and the docs are now updated.

hadley:
Thanks!

PR #227, opened 2021-03-19
Translate `desc(col)` to `-col` inside `arrange()`

markfairbanks:


hadley:
And that works for character vectors?

markfairbanks:
Yep. It's their recommended way so that you can sort character vectors descending using their radix sorting.

``` r
library(data.table)
library(dplyr)

dt <- data.table(x = sample(letters, 100000, TRUE))

bench::mark(
  minus = dt[order(-x)],
  desc = dt[order(desc(x))],
)
#> # A tibble: 2 x 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 minus        1.16ms   1.71ms    543.      2.85MB     13.3
#> 2 desc       437.11ms 450.29ms      2.22    8.06MB      0
```

hadley:
Good to know, thanks!

PR #234, opened 2021-03-29
Translated named args in `if_else()`

markfairbanks:
reprex of issue fixed:

``` r
library(data.table)
library(dplyr)
library(dtplyr)

test_df <- data.table(x = c("a", "a", "b", NA))

test_df %>%
  mutate(if_x = if_else(x == "a", true = 1, false = 0, missing = NULL))
#> Error in fifelse(x == "a", true = 1, false = 0, missing = NULL): unused arguments (true = 1, false = 0, missing = NULL)
```

### DIFF
--- a/R/tidyeval.R
+++ b/R/tidyeval.R
@@ -104,7 +104,7 @@ dt_squash_call <- function(x, env, data, j = TRUE) {
     } else if (is_symbol(x[[2]], ".env")) {
       sym(paste0("..", var))
     }
-  } else if (is_call(x, "coalesce") || is_call(x, "replace_na")) {
+  } else if (is_call(x, c("coalesce", "replace_na"))) {
     x[[1L]] <- quote(fcoalesce)
     x
   } else if (is_call(x, "case_when")) {
@@ -133,7 +133,7 @@ dt_squash_call <- function(x, env, data, j = TRUE) {
     x[[1]] <- sym("-")
     x[[2]] <- get_expr(x[[2]])
     x
-  } else if (is_call(x, "if_else") || is_call(x, "ifelse")) {
+  } else if (is_call(x, c("if_else", "ifelse"))) {
     x[[1L]] <- quote(fifelse)
     names(x) <- if_else_args[names(x)]
     x[-1] <- lapply(x[-1], dt_squash, env, data, j = j)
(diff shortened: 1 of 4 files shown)


R/tidyeval.R
```diff
@@ -133,8 +133,9 @@ dt_squash_call <- function(x, env, data, j = TRUE) {
     x[[1]] <- sym("-")
     x[[2]] <- get_expr(x[[2]])
     x
-  } else if (is_call(x, "if_else") || is_call(x, "ifelse")) {
+  } else if (is_call(x, c("if_else", "ifelse"))) {
     x[[1L]] <- quote(fifelse)
+    names(x) <- if_else_args[names(x)]
```
hadley, line 138: I wonder if it's worth using `match.call()` here — the small advantage is that it would protect against partial naming; the bigger disadvantage is that it would turn unnamed arguments to named arguments.

markfairbanks: Maybe use `match.call()` and then remove names from unnamed arguments?

``` r
library(rlang)

if_else_args <- c(
  "condition" = "test", "true" = "yes",
  "false" = "no", "missing" = "na"
)

x <- expr(if_else(x < 1, tr = 1, 0))

if (is_call(x, c("if_else", "ifelse"))) {
  named_flag <- have_name(x)
  if (any(named_flag)) {
    if (is_call(x, "if_else")) {
      x <- match.call(dplyr::if_else, x)
      names(x) <- if_else_args[names(x)]
    } else {
      x <- match.call(base::ifelse, x)
    }
    names(x)[!named_flag] <- ""
  }
  
  x[[1]] <- quote(fifelse)
  x
}
#> fifelse(x < 1, yes = 1, 0)
```

hadley: I don't think that works because `match.call()` also changes the argument order:

``` r
x <- rlang::expr(if_else(x < 1, 0, tr = 1))
match.call(dplyr::if_else, x)
#> if_else(condition = x < 1, true = 1, false = 0)
```

How about just unconditionally stripping the argument names in the generated call?

markfairbanks: Sounds good - seems like a good tradeoff

markfairbanks:
All set for review

PR #241, opened 2021-05-07
Add `step_setnames()`

mgirlich:
`step_call("setnames")` is used multiple times in the code, so it might as well be replaced with its own step.

### DIFF
--- a/R/step-call-pivot_wider.R
+++ b/R/step-call-pivot_wider.R
@@ -128,13 +128,13 @@ pivot_wider.dtplyr_step <- function(data,
   }
 
   if (!is.null(names_glue)) {
-    out <- step_setnames(out, !!!stats::setNames(new_vars, glue_vars), in_place = FALSE)
+    out <- step_setnames(out, new_vars, glue_vars, in_place = FALSE)
 
     # In case of names_sort = TRUE
     new_vars <- glue_vars
   } else if (nchar(names_prefix) > 0) {
     new_names <- paste0(names_prefix, new_vars)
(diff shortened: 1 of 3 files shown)


R/step-call-pivot_wider.R
```diff
@@ -128,26 +128,13 @@ pivot_wider.dtplyr_step <- function(data,
...
 
   if (!is.null(names_glue)) {
-    out <- step_call(
-      out,
-      "setnames",
-      args = list(old = new_vars, new = glue_vars),
-      vars = c(id_cols, glue_vars),
-      in_place = FALSE
-    )
+    out <- step_setnames(out, !!!stats::setNames(new_vars, glue_vars), in_place = FALSE)
```
hadley, line 131: Do you think it would be easier to understand if we turned `new_vars` and `old_vars` into explicit arguments?

R/step-setnames.R
```diff
@@ -0,0 +1,37 @@
+step_setnames <- function(x, ..., in_place, rename_groups = FALSE) {
+  stopifnot(is_step(x))
+  stopifnot(is_bool(in_place))
+  stopifnot(is_bool(rename_groups))
+
+  sim_data <- simulate_vars(x)
+  locs <- tidyselect::eval_rename(expr(c(...)), sim_data)
```
hadley, line 7: This logic wasn't previously applied in (e.g.) `pivot_wider.dtplyr_step()`. Is it likely to cause problems?

hadley:
Thanks!

PR #276, opened 2021-07-28
Translate `n_distinct()`

markfairbanks:


R/tidyeval.R
```diff
@@ -145,6 +145,19 @@ dt_squash_call <- function(x, env, data, j = TRUE) {
     x
   } else if (is_call(x, "n", n = 0)) {
     quote(.N)
+  } else if (is_call(x, "n_distinct")) {
+    x <- match.call(dplyr::n_distinct, x, expand.dots = FALSE)
```
hadley, line 149: Maybe use `call_match()` (from rlang) just for consistency?

R/tidyeval.R
```diff
@@ -145,6 +145,19 @@ dt_squash_call <- function(x, env, data, j = TRUE) {
...
+    dots <- x$...
+    if (length(dots) == 1) {
+      vec <- dots[[1]]
+    } else {
+      vec <- call2("c", !!!dots)
```
hadley, line 154: Should this create a data.table? I think that's truer to `n_distinct()`:

``` r
dplyr::n_distinct(c(1, 1, 2), c(1, 2, 1))
#> [1] 3
data.table::uniqueN(c(c(1, 1, 2), c(1, 2, 1)))
#> [1] 2
data.table::uniqueN(data.table::data.table(c(1, 1, 2), c(1, 2, 1)))
#> [1] 3
```

<sup>Created on 2021-07-30 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

R/tidyeval.R
markfairbanks: Looks like `call_match()` is new to the dev version of rlang. Should I add the dev version of rlang as a dependency?

And if yes, should I switch all of the other `match.call()` uses over to `call_match()` in this PR?

R/tidyeval.R
markfairbanks: Interesting, I didn't know that was the behavior of `n_distinct()` with multiple inputs. Yep, I can switch this over.

R/tidyeval.R
hadley: Oh if it's only in dev version, don't worry about it.

markfairbanks:
All set for review

PR #279, opened 2021-08-09
Make sure y is a `lazy_dt` in set ops

### LINKED ISSUE #278: `setdiff` results in error
Using `setdiff` causes an error

``` r
a <- data.table(x = 1:2)
b <- data.table(x = 2)
setdiff(a, b)
#>    x
#> 1: 1

setdiff(a, b)
#> Error in step_set(x, y, style = "setdiff"): is_step(y) is not TRUE
```

<sup>Created on 2021-08-09 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

markfairbanks:
Closes #278

### DIFF
--- a/R/step-set.R
+++ b/R/step-set.R
@@ -23,10 +23,10 @@ dt_call.dtplyr_step_set <- function(x, needs_copy = x$needs_copy) {
   rhs <- dt_call(x$parent2)
 
   call <- switch(x$style,
-                 intersect = call2("fintersect", lhs, rhs),
-                 union = call2("funion", lhs, rhs),
-                 union_all = call2("funion", lhs, rhs, all = TRUE),
-                 setdiff = call2("fsetdiff", lhs, rhs),
+    intersect = call2("fintersect", lhs, rhs),
+    union = call2("funion", lhs, rhs),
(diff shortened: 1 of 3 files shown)


R/step-set.R
```diff
@@ -23,10 +23,10 @@ dt_call.dtplyr_step_set <- function(x, needs_copy = x$needs_copy) {
   rhs <- dt_call(x$parent2)
 
   call <- switch(x$style,
-    intersect = call2("fintersect", lhs, rhs),
-    union = call2("funion", lhs, rhs),
-    union_all = call2("funion", lhs, rhs, all = TRUE),
-    setdiff = call2("fsetdiff", lhs, rhs),
+                 intersect = call2("fintersect", lhs, rhs),
```
hadley, line 26: Do you mind leaving the indenting as is here? Otherwise looks good.

hadley (approved):

R/step-set.R
markfairbanks: Whoops my bad, not sure how I even ended up changing that

hadley:
(In cases like this you're welcome to make the change and merge; if I've approved the PR it means I don't need to see it again)

PR #285, opened 2021-08-22
desc() checks the number of arguments.

eutwt:
Currently if you pass two arguments to `desc` as in `desc(a, b)`, it is converted to `a - b` by `dt_squash_call`. This PR adds a check to error if more than one argument is provided. 

A similar issue was fixed in dplyr: https://github.com/tidyverse/dplyr/pull/5929

Before this PR:

``` r

dt <- lazy_dt(data.frame(a = rep(1:2, each = 2), b = 1:4))

dt %>% 
  arrange(desc(a, b))
#> Source: local data table [4 x 2]
#> Call:   `_DT1`[order(a - b)]
#> 
#>       a     b
#>   <int> <int>
#> 1     2     4
#> 2     1     2
#> 3     2     3
#> 4     1     1
#> 
```

<sup>Created on 2021-08-22 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

After this PR:

``` r

dt <- lazy_dt(data.frame(a = rep(1:2, each = 2), b = 1:4))

dt %>% 
  arrange(desc(a, b))
#> Error: `desc()` expects exactly one argument.
```

<sup>Created on 2021-08-22 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,10 +1,6 @@
 # dtplyr (development version)
 
-* `tally()` and `count()` now follow the dplyr convention of creating a unique 
-  name if the default output `name` (n) already exists (@eutwt, #295).
-
-* Fixed a bug which prevented changing the value of a group variable with
-  `summarise()`, `tally()`, or `count()` (@eutwt, #295).
+* `desc()` now gives an error if more than one argument is provided (@eutwt #285).
 
 * `if_any()` and `if_all()` now default to `.cols = everything()` when `.cols` 
   isn't provided (@eutwt, #294).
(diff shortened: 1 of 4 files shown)


hadley:
Looks good — can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

eutwt:
This is fixing a bug in a feature which has not yet been in any release, and the feature itself already has a bullet. Should I still add a new one?

eutwt:
I went ahead and added one, but can also revert the commit, either way. Not sure what the convention is.

hadley:
Oh, support for `desc()` is new in this release? In that case, no bullet is necessary. Sorry!

eutwt:
Yep, new this release. Reverted the news commit

hadley:
Thanks!

Issue #287, opened 2021-08-27
`mutate` inconsistencies with dplyr


eutwt:
Using this issue to maintain a collection of (mostly minor) differences between dplyr and dtplyr mutate. Maybe it's easier to prioritize between them with them all listed together, if there's a desire to fix them.

Please don't take this as criticism of dtplyr, I love using this package  😄

1. ~~Unusual `across` results when `.fns` is a formula whose RHS is not a call~~
    -  covered by #288 
2. ~~Cannot use anonymous (non-purrr) functions in `across`~~
    -  covered by #288 
3. Cannot use purrr-style formula functions with more than 1 argument
4. #281

5. Cannot remove new columns created in mutate with `[colnm] = NULL`
6. ~~Cannot use `.before` and `.after`~~
    - covered by #291
7. Cannot use `.keep`
8. No splicing of data.frame inputs
9. Cannot treat across output as a data.frame e.g. `df %>% mutate(a = do.call(coalesce, across(starts_with('a'))))`

------
**3. Cannot use purrr-formula functions with more than 1 argument**

``` r
library(dplyr, warn.conflicts = FALSE)
library(data.table, warn.conflicts = FALSE)

dt <- data.table(a = 1:2)

dt %>%
  mutate(across(a, ~ .x + .y, 1)) %>% 
  as_tibble()
#> # A tibble: 2 × 1
#>       a
#>   <dbl>
#> 1     2
#> 2     3

library(dtplyr)

dt %>%
  mutate(across(a, ~ .x + .y, 1)) %>% 
  as_tibble()
#> Error in eval(jsub, SDenv, parent.frame()): object '.y' not found
```

<sup>Created on 2021-08-27 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

**5. Cannot remove new columns created in mutate with `[colnm] = NULL`**

``` r
library(dplyr, warn.conflicts = FALSE)
library(data.table, warn.conflicts = FALSE)

dt <- data.table(a = 1:2)

dt %>%
  mutate(b = 3, b = NULL) %>% 
  as_tibble()
#> # A tibble: 2 × 1
#>       a
#>   <int>
#> 1     1
#> 2     2

library(dtplyr)

dt %>%
  mutate(b = 3, b = NULL) %>% 
  as_tibble()
#> Error in `[.data.table`(copy(`_DT1`), , `:=`(b = 3, b = NULL)): Can't assign to the same column twice in the same query (duplicates detected).
```

[...]

PR #288, opened 2021-08-27
Allow functions and formulas with non-call RHS in `across`

eutwt:
This PR does two things. Included together since they're both expanding possible `across` inputs and both pretty small changes.

1. Skips `dt_squash_call` for formulas whose RHS is not a call, allowing such formulas in `across`
- Behavior of non-call formulas in master shown here: #287  
2. Allows anonymous non-formula functions in across

I could separate these if needed (and if there's interest in accepting either of them)

``` r

lazy_dt(data.frame(a = 1, b = 2)) %>% 
  mutate(across(a, ~ b)) %>% 
  as_tibble()
#> # A tibble: 1 × 2
#>       a     b
#>   <dbl> <dbl>
#> 1     2     2

lazy_dt(data.frame(a = 1, b = 2)) %>% 
  mutate(across(a:b, function(x) x + 1)) %>% 
  as_tibble()
#> # A tibble: 1 × 2
#>       a     b
#>   <dbl> <dbl>
#> 1     2     3
```

<sup>Created on 2021-08-27 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/tests/testthat/test-tidyeval-across.R
+++ b/tests/testthat/test-tidyeval-across.R
@@ -58,6 +58,17 @@ test_that("across() translates functions", {
   )
 })
 
+test_that("across() captures anonymous functions", {
+  dt <- lazy_dt(data.frame(a = 1))
+
+  expect_equal(
+   capture_across(dt, across(a, function(x) log(x))),
+   quo_squash(list(
+     a = call2(function(x) log(x), quote(a))
+   ))
+  )
+})
+
 test_that("dots are translated too", {
   fun <- function() {
     dt <- lazy_dt(data.frame(a = 1, b = 2))
--- a/tests/testthat/test-tidyeval-across.R
+++ b/tests/testthat/test-tidyeval-across.R
@@ -63,9 +63,7 @@ test_that("across() captures anonymous functions", {
 
   expect_equal(
    capture_across(dt, across(a, function(x) log(x))),
-   quo_squash(list(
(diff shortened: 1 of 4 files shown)


tests/testthat/test-tidyeval-across.R
```diff
@@ -58,6 +58,15 @@ test_that("across() translates functions", {
   )
 })
 
+test_that("across() captures anonymous functions", {
+  dt <- lazy_dt(data.frame(a = 1))
+
+  expect_equal(
+   capture_across(dt, across(a, function(x) log(x))),
+   quo_squash(list(a = call2(function(x) log(x), quote(a))))
```
hadley, line 66: Could you construct the expected value a little more directly? I don't remember what `quo_squash()` does off the top of my head.

hadley (commented):
Looks good — just one question and can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

tests/testthat/test-tidyeval-across.R
eutwt: Oops, it was a mistake to have it there, thanks for catching that. In this case it wasn't doing anything. I've removed it.

eutwt:
Should be all set now

PR #291, opened 2021-08-29
Add `.before` and `.after` arguments to `mutate`

eutwt:
Example with this PR loaded:

``` r
library(dtplyr)

lazy_dt(data.frame(x = 1, y = 1)) %>% 
  mutate(z = 1, .before = y)
#> Call:   setcolorder(copy(`_DT1`)[, `:=`(z = 1)], c("x", "z", "y"))
#> 
#>       x     z     y
#>   <dbl> <dbl> <dbl>
#> 1     1     1     1
#> 
```

<sup>Created on 2021-08-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/tests/testthat/test-step-mutate.R
+++ b/tests/testthat/test-step-mutate.R
@@ -105,3 +105,27 @@ test_that("emtpy mutate returns input", {
   expect_equal(mutate(dt), dt)
   expect_equal(mutate(dt, !!!list()), dt)
 })
+
+# .before and .after -----------------------------------------------------------
+
+test_that("can use .before and .after to control column position", {
+  dt <- lazy_dt(data.frame(x = 1, y = 2))
+  expect_named(
+    mutate(dt, z = 1) %>% as_tibble(),
+    c("x", "y", "z")
+  )
+  expect_named(
+    mutate(dt, z = 1, .before = x) %>% as_tibble(),
+    c("z", "x", "y")
+    )
+  expect_named(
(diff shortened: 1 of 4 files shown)


tests/testthat/test-step-mutate.R
```diff
@@ -105,3 +105,27 @@ test_that("emtpy mutate returns input", {
...
+test_that("can use .before and .after to control column position", {
+  dt <- lazy_dt(data.frame(x = 1, y = 2))
+  expect_named(
+    mutate(dt, z = 1) %>% as_tibble(),
+    c("x", "y", "z")
+  )
+  expect_named(
+    mutate(dt, z = 1, .before = x) %>% as_tibble(),
+    c("z", "x", "y")
+    )
```
hadley, line 120: Indenting is slightly off here and below.

hadley (commented):
In general, I don't think dtplyr needs to provide experimental arguments, but I'm pretty sure that all the current dplyr experimental arguments are going to stay, so this still makes sense. Otherwise looks good, just one small styling suggestion and it needs a news bullet 😄

tests/testthat/test-step-mutate.R
eutwt: Thanks, fixed.

eutwt:
Should be all set now

hadley:
BTW we add news bullets to the top, since it makes resolving the merge conflicts a little easier.

PR #293, opened 2021-08-31
`across()` handles named selections

eutwt:
Should work the same as dplyr, as implemented in https://github.com/tidyverse/dplyr/pull/5889.

``` r
library(dtplyr)

lazy_dt(data.frame(x = 1, y = 2)) %>% 
  summarise(across(c(z = y), ~ .^2))
#> Call:   `_DT1`[, .(z = y^2)]
#> 
#>       z
#>   <dbl>
#> 1     4
#> 
```

<sup>Created on 2021-08-31 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Things to note:

- Allowing named selections makes empty selections possible, so the empty-selection case for mutate is also handled in this PR.
- The "non-expanded" cases tested in the dplyr PR don't work in either the current GitHub dtplyr or this PR.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # dtplyr (development version)
 
-* `across()` handles named selections (@eutwt #293)
+* `across()` handles named selections (@eutwt #293).
 
 * `summarise()` now supports the `.groups` argument (@mgirlich, #245).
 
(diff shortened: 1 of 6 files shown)


hadley:
Thanks!

PR #295, opened 2021-09-04
Avoid duplicated columns in `count`, `tally`, and `summarise`

eutwt:
This fixes the following errors (and instead produces results matching dplyr)

``` r
library(dtplyr)
dt <- lazy_dt(data.frame(a = 2, n = 3))

dt %>% 
  count(a, name = 'a') %>% 
  as_tibble()
#> Error: Column name `a` must not be duplicated.
#> Use .name_repair to specify repair.

dt %>% 
  group_by(a) %>% 
  summarise(a = 1) %>% 
  as_tibble()
#> Error: Column name `a` must not be duplicated.
#> Use .name_repair to specify repair.

dt %>% 
  count(n) %>% 
  as_tibble()
#> Error: Column name `n` must not be duplicated.
#> Use .name_repair to specify repair.
```

<sup>Created on 2021-09-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # dtplyr (development version)
 
+* `tally()` and `count()` now follow the dplyr convention of creating a unique 
+  name if the default output `name` (n) already exists (@eutwt, #295).
+
+* Fixed a bug which prevented changing the value of a group variable with
+  `summarise()`, `tally()`, or `count()` (@eutwt, #295).
+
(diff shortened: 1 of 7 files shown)


eutwt:
I gave a name to an input in one of the pivot_wider tests because earlier the snapshot of the expression was showing a different name, e.g. ``setnames(dcast(`_DT17`...`` instead of ``setnames(dcast(`_DT12`...``

hadley:
Thanks!

PR #296, opened 2021-09-04
Translate `semi_join` more directly using `mult = 'first'`

eutwt:
This translates `semi_join` to data.table is more directly. Should be a little faster too.

``` r
library(rlang)

a <- data.table(V1 = sample(3e6))[, V2 := 1]
b <- data.table(V1 = c(-sample(1e3), sample(2e6)))[, V2 := 2]
# give b many cols to make sure pr doesn't copy rhs cols
b[, paste0('V',3:40) := as.list(3:40)]

curr <- function(lhs, rhs, by) {
  call <- call2("[", lhs, call2("unique", call2("[", lhs, rhs, which = TRUE, nomatch = NULL, on = by)))
  eval(call)
}

pr <- function(lhs, rhs, by) {
  vars_lhs <- names(lhs)
  call <- call2("[", lhs, rhs, j = vars_lhs, on = by, nomatch = NULL, mult = 'first')
  eval(call)
}

library(bench)
mark(
  curr(a, b, by = 'V1'),
  pr(a, b, by = 'V1')
)
#> Warning: Some expressions had a GC in every iteration; so filtering is disabled.
#> # A tibble: 2 × 6
#>   expression                 min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>            <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 curr(a, b, by = "V1")    432ms    441ms      2.27     127MB     3.40
#> 2 pr(a, b, by = "V1")      338ms    342ms      2.92     111MB     0
```

<sup>Created on 2021-09-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # dtplyr (development version)
 
+* `semi_join()` is now translated as a single `[.data.table` call with `nomatch = NULL`
+  and `mult = 'first'` (@eutwt #296)
+
 * `if_any()` and `if_all()` now default to `.cols = everything()` when `.cols` 
   isn't provided (@eutwt, #294).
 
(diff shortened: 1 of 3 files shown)


eutwt:
Closing because this does not actually do a semi-join if the lhs is not unique on the join key

PR #305, opened 2021-10-05
dt_sqash call args for `coalesce` and `replace_na`

### LINKED ISSUE #304: replace_na() can't find local variable
``` r
library(dtplyr)

dt <- lazy_dt(data.frame(x = c(1, NA)), "dt")

local({
  x_miss <- 2
  
  dt %>% mutate(x = tidyr::replace_na(x, x_miss)) 
})
#> Error in fcoalesce(x, x_miss): object 'x_miss' not found
```

<sup>Created on 2021-10-05 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Presumably because we're not correctly inlining `x_miss` into the call as we normally do:

``` r
local({
  x_miss <- 2
  
  dt %>% mutate(x = x + x_miss) %>% show_query()
})
#> copy(dt)[, `:=`(x = x + 2)]
```

<sup>Created on 2021-10-05 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

@markfairbanks does this ring any bells?

eutwt:
I believe this closes #304. I just applied the code used for `case_when` to the `coalesce`/`replace_na` section.


hadley:
Thanks for the fix! I think `lag`/`lead` might need it too?

hadley:
Actually let me finish it off since I have couple of minutes free.

eutwt:
Sure. Looks like `desc` and `n_distinct` need it too

PR #308, opened 2021-10-06
Allow Cartesian joins in inner joins

hadley:
For consistency with other types


hadley:
@mgirlich looks like it doesn't break anything significant

PR #315, opened 2021-12-03
Add @eutwt to authors

eutwt:
Thanks :)


eutwt:
I think the test failures are unrelated 

https://github.com/tidyverse/dtplyr/runs/4409047875?check_suite_focus=true#step:5:112

hadley:
It'd be concerning if they were 😆

PR #320, opened 2021-12-08
Allow "temp" vars in mutate by adding var = NULL at end

eutwt:
So that this works:

``` r
library(dtplyr)
  mutate(
    y = 2,
    z = y*2,
    y = NULL
  )
#> Source: local data table [1 x 2]
#> Call:   copy(`_DT1`)[, `:=`(c("z"), {
#>     y <- 2
#>     z <- y * 2
#>     .(z)
#> })]
#> 
#>       x     z
#>   <dbl> <dbl>
#> 1     1     4
#> 
```

<sup>Created on 2021-12-07 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


eutwt:
Closing because I think #323 is a better approach

PR #323, opened 2021-12-10
rework capture_dots to progressively track vars

### LINKED ISSUE #281: Cannot access previous columns from `mutate` using `across`
``` r
dt <- data.table(y = 2)
dt %>%
  mutate(z = 4, 
         across(z, sqrt))
#>    y z
#> 1: 2 2

dt %>%
  mutate(z = 4, 
         across(z, sqrt))
#> Error: Can't subset columns that don't exist.
#> x Column `z` doesn't exist.
```

<sup>Created on 2021-08-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### LINKED ISSUE #321: Cannot use nested variables in mutate if variable name matches object in global environment
This should work but doesn't.

``` r
library(dtplyr)

  mutate(y = 2, z = y*2)
#> Call:   copy(`_DT1`)[, `:=`(c("y", "z"), {
#>     y <- 2
#>     z <- y * 2
#>     .(y, z)
#> })]
#> 
#>       x     y     z
#>   <dbl> <dbl> <dbl>
#> 1     1     2     4
#> 

y <- 'global var'

  mutate(y = 2, z = y*2)
#> Error in ..y * 2: non-numeric argument to binary operator

  mutate(y = 2, z = y*2) %>% 
  show_query()
#> copy(`_DT3`)[, `:=`(y = 2, z = ..y * 2)]
```

<sup>Created on 2021-12-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

eutwt:
closes #321
closes #281
closes #320

Because `capture_dots` in `mutate.dtplyr_step` does not keep track of new variables as they are created, the four actions below result in an error on the main branch. This PR also fixes the analogous actions for `transmute`

``` r
library(dtplyr)

dt <- lazy_dt(data.frame(x = 1))
mutate(dt, y = 2, z = y*2, y = NULL)
#> Source: local data table [1 x 2]
#> Call:   copy(`_DT1`)[, `:=`(c("z"), {
#>     y <- 2
#>     z <- y * 2
#>     .(z)
#> })]
#> 
#>       x     z
#>   <dbl> <dbl>
#> 1     1     4
#> 
mutate(dt, y = 2, across(y, sqrt))
#> Source: local data table [1 x 2]
#> Call:   copy(`_DT1`)[, `:=`(c("y"), {
#>     y <- 2
#>     y <- sqrt(y)
#>     .(y)
#> })]
#> 
#>       x     y
#>   <dbl> <dbl>
#> 1     1  1.41
#> 
mutate(dt, y = 2, y = 3)
#> Source: local data table [1 x 2]
#> Call:   copy(`_DT1`)[, `:=`(c("y"), {
#>     y <- 2
#>     y <- 3
#>     .(y)
#> })]
#> 
#>       x     y
#>   <dbl> <dbl>
#> 1     1     3
#> 
y <- 'global var'
mutate(dt, y = 2, z = y + 1)
#> Call:   copy(`_DT1`)[, `:=`(c("y", "z"), {
#>     y <- 2
#>     z <- y + 1
#>     .(y, z)
#> })]
#> 
#>       x     y     z
#>   <dbl> <dbl> <dbl>
#> 1     1     2     3
#> 
```

<sup>Created on 2021-12-10 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/R/tidyeval.R
+++ b/R/tidyeval.R
@@ -49,7 +49,8 @@ capture_new_vars <- function(.data, ...) {
     } else {
       dots[[i]] <- dot
     }
-    .data$vars <- union(.data$vars, names(dots)[i])
+    new_vars <- names(dot) %||% names(dots)[i]
+    .data$vars <- union(.data$vars, new_vars)
   }
 
   # Remove names from any list elements
--- a/R/tidyeval.R
+++ b/R/tidyeval.R
@@ -44,7 +44,11 @@ capture_new_vars <- function(.data, ...) {
   dots <- as.list(enquos(..., .named = TRUE))
   for (i in seq_along(dots)) {
     dot <- dt_squash(dots[[i]], data = .data) 
-    if (!is.null(dot)) dots[[i]] <- dot
+    if (!is.null(dot)) {
+      dots[[i]] <- dot
+    } else {
+      dots[i] <- list(NULL)
+    }
     .data$vars <- union(.data$vars, names(dots)[i])
   }
 
@@ -57,7 +61,7 @@ capture_new_vars <- function(.data, ...) {
   out <- unlist(dots, recursive = FALSE)
 
   # detect temp vars i.e. repeated vars where last is var = NULL
-  var_is_null <- vapply(out, identical, y = quo(NULL), lgl(1))
+  var_is_null <- vapply(out, is.null, lgl(1))
   if (any(var_is_null)) {
     is_repeated <- duplicated(names(out))
     is_last <- !duplicated(names(out), fromLast = TRUE)
--- a/R/tidyeval.R
+++ b/R/tidyeval.R
@@ -41,7 +41,7 @@ capture_dots <- function(.data, ..., .j = TRUE) {
 }
 
 capture_new_vars <- function(.data, ...) {
-  dots <- enquos(..., .named = TRUE)
+  dots <- as.list(enquos(..., .named = TRUE))
   for (i in seq_along(dots)) {
     dot <- dt_squash(dots[[i]], data = .data) 
     dots[[i]] <- dot %||% structure('zap', class = 'rlang_zap')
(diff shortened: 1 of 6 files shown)


eutwt:
Need to go back and review the code to see where that rlang warning is coming from

R/tidyeval.R
```diff
@@ -40,6 +40,37 @@ capture_dots <- function(.data, ..., .j = TRUE) {
...
+  names(dots)[is_list] <- ""
+
+  # Auto-splice list results from dt_squash()
+  dots[!is_list] <- lapply(dots[!is_list], list)
+  out <- unlist(dots, recursive = FALSE)
+
+  # detect temp vars i.e. repeated vars where last is var = NULL
+  var_is_null <- vapply(out, identical, y = quo(NULL), lgl(1))
+  if (any(var_is_null)) {
+    var_is_last <- rev(data.table::rowid(rev(names(out))) == 1)
```
eutwt, line 62: Another option is below. Neither of these seem good.
`!duplicated(match(names(out), rev(names(out))), fromLast = TRUE)`

eutwt (commented):
Meant to just leave a comment, not start review

eutwt:
@markfairbanks would it make more sense for the tests to check the expressions or the data frame outputs here?

markfairbanks:
> would it make more sense for the tests to check the expressions or the data frame outputs here?

I would probably add some tests to check the expressions to the current test set you've already added. Can't hurt to have both.

Here's an example of [some that I wrote for `separate()`](https://github.com/tidyverse/dtplyr/blob/main/tests/testthat/test-step-subset-separate.R). It has some data frame/value tests and also has some expression tests.

eutwt:
Apologies for asking you to review and then closing @markfairbanks, but I'm going to close this because it doesn't account for the case where `var = NULL` is provided and the query is nested, e.g. below (which also doesn't work on the main branch, but still). I think to account for all the edge cases like this an extra `[` step for var-removal is needed, which is a substantially different approach than this PR currently so I'll close this and open a new one. 

```
dt <- lazy_dt(data.frame(x = 1))
step <- dt %>% mutate(x = 2, z = x*2, x = NULL) 
```

PR #324, opened 2021-12-12
mutate edge cases

### LINKED ISSUE #281: Cannot access previous columns from `mutate` using `across`

### LINKED ISSUE #321: Cannot use nested variables in mutate if variable name matches object in global environment

eutwt:
closes #321
closes #281

All the tests added to test-step-mutate.R result in an error on the main branch. These are all pretty rare use-cases, but trying to make dtplyr as compatible with existing working dplyr code as possible, even if it's a bit odd.

The changes are:

- Solved by `process_new_vars`
  - If there are **repeated variable names**, e.g. `mutate(dt, z = 2, z = 3)` this needs to be handled the same way nested arguments are handled on the main branch (even if there is no nesting) to avoid an error. `mutate_nested_vars` is renamed to `mutate_with_braces` because it's= now not only used when arguments are nested.
    - [test](https://github.com/tidyverse/dtplyr/pull/324/files#diff-68dfa5e2cb0290c32457805c202b1e388a6823e6dabaeacaa3f9f4fd13f4ef13R73)
  - If there is a **variable removal**, i.e. a mutate argument `var = NULL` not followed by any other mutate arguments named `var` (e.g. not `var = NULL, var = 3`), then that variable should not be in the final output.
    - If the data is grouped, you need an additional step (`[.data.table` call) to remove the variable because `DT[, var := NULL, by = g]` results in an error.
    - If we are in a `mutate_with_braces` branch, it's hard to determine how you should modify the current `j` expression to get the correct output. Some things that seem like they should work don't, see example below. So to keep things simple, this case is handled the same as grouped data, i.e. an extra `[.data.table` call is added.
    - [test](https://github.com/tidyverse/dtplyr/pull/324/files#diff-68dfa5e2cb0290c32457805c202b1e388a6823e6dabaeacaa3f9f4fd13f4ef13R159)
 
 ```{r}
library(data.table)

dd <- data.table(x = 1)
dd[, c('z', 'x') := {x <- NULL; list(2, NULL)}][]
#>        x     z
#>    <num> <num>
#> 1:     2     2
```

<sup>Created on 2021-12-23 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

- Solved by `capture_new_vars`
  - The arguments need to be captured sequentially, so that the list of previously created variables is accessible at each capture. This fixes an issue with conflicting global variables and not being able to access previously created variables in mutate. 
    - [test](https://github.com/tidyverse/dtplyr/pull/324/files#diff-68dfa5e2cb0290c32457805c202b1e388a6823e6dabaeacaa3f9f4fd13f4ef13R120)

### DIFF
--- a/R/step-mutate.R
+++ b/R/step-mutate.R
@@ -76,6 +76,8 @@ mutate_with_braces <- function(mutate_vars) {
 mutate.dtplyr_step <- function(.data, ...,
                                .before = NULL, .after = NULL) {
   dots <- capture_new_vars(.data, ...)
+  trivial_dot <- imap(dots, ~ is_symbol(.x) && sym(.y) == .x && .y %in% .data$vars)
+  dots <- dots[!as.vector(trivial_dot, "logical")]
   dots_list <- process_new_vars(.data, dots)
   dots <- dots_list$dots
   if (is_null(dots) || is_empty(dots)) {
--- a/R/step-mutate.R
+++ b/R/step-mutate.R
@@ -27,29 +27,17 @@ dt_call.dtplyr_step_mutate <- function(x, needs_copy = x$needs_copy) {
 
   out <- call2("[", dt_call(x$parent, needs_copy), , j)
 
-  if (x$use_braces && !is_empty(mutate_list$removed_vars)) {
-    out <- call2("[", out, , call2(":=", mutate_list$removed_vars, NULL))
-  }
-
   add_grouping_param(out, x, arrange = FALSE)
 }
 
 mutate_with_braces <- function(mutate_vars) {
-  # removing vars with var = NULL doesn't work well with expressions using braces
-  # need to identify those cases to deal with them in a separate step
-  var_is_null <- vapply(mutate_vars, is.null, lgl(1))
-  is_last <- !duplicated(names(mutate_vars), fromLast = TRUE)
-  var_removal <- var_is_null & is_last
-  mutate_vars <- mutate_vars[!var_removal]
-
   assign <- map2(syms(names(mutate_vars)), mutate_vars, function(x, y) call2("<-", x, y))
   new_vars <- unique(names(mutate_vars))
   output <- call2(".", !!!syms(new_vars))
 
   list(
     expr = call2("{", !!!assign, output),
-    new_vars = new_vars,
-    removed_vars = names(var_removal)[var_removal]
+    new_vars = new_vars
   )
 }
 
@@ -86,11 +74,22 @@ mutate_with_braces <- function(mutate_vars) {
 mutate.dtplyr_step <- function(.data, ...,
                                .before = NULL, .after = NULL) {
   dots <- capture_new_vars(.data, ...)
-  if (is_null(dots)) {
+  if (is_empty(dots)) {
     return(.data)
   }
 
-  use_braces <- nested_vars(.data, dots, .data$vars) | anyDuplicated(names(dots))
+  var_removals <- vapply(dots, is_var_removal, logical(1))
+  vars_removed <- names(var_removals)[var_removals]
+  nested_vars <- nested_vars(.data, dots, .data$vars)
+  repeated_vars <- anyDuplicated(names(dots))
+  use_braces <- nested_vars | repeated_vars
+  grouped_data <- !is_empty(.data$groups)
+  need_removal_step <- any(var_removals) && (use_braces | grouped_data)
+  if (need_removal_step) {
+    dots <- dots[!var_removals]
+  } else {
+    dots <- unmark_var_removals(dots, var_removals)  
+  }
   out <- step_mutate(.data, dots, use_braces)
 
   .before <- enquo(.before)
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: R/step-mutate.R [hunk 1]
```diff
@@ -1,6 +1,8 @@
-step_mutate <- function(parent, new_vars = list(), nested = FALSE) {
+step_mutate <- function(parent, new_vars = list(), use_braces = FALSE) {
   vars <- union(parent$vars, names(new_vars))
-  vars <- setdiff(vars, names(new_vars)[vapply(new_vars, is_null, lgl(1))])
+  var_is_null <- vapply(new_vars, is_null, lgl(1)) 
+  is_last <- !duplicated(names(new_vars), fromLast = TRUE)
```
eutwt, line 4: Need this so that `mutate(dt, y = NULL, y = 3)$vars` contains `"y"`

eutwt:
Would appreciate if you could take a look when you get a chance @markfairbanks , but no rush at all considering it's the holiday season and the only issues this closes are those opened by me :)

markfairbanks:
@eutwt I'll take a look at this in the next couple days 😄 Hope you had a good holiday season.

eutwt:
Thanks for the review :)

mgirlich:
@eutwt Maybe you want to add one more minor improvement:

``` r

lazy_dt(mtcars) %>% 
  mutate(mpg) %>% 
  show_query()
#> copy(`_DT1`)[, `:=`(mpg = mpg)]
```

<sup>Created on 2022-02-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Unlikely that a user does that but this can happen e.g. in functions (I stumbled over it when I wanted to use `dplyr::group_by_prepare()` in `group_by.dtplyr_step()`). I think the following should work

```r
  dots <- capture_new_vars(.data, ...)
  trivial_dot <- purrr::imap_lgl(dots, ~ is_symbol(.x) && sym(.y) == .x)
  dots <- purrr::discard(dots, trivial_dot)
```

hadley:
@markfairbanks sorry I missed this earlier, but I generally don't need to review unless there's something specific that you want me to look at.

markfairbanks:
@mgirlich I like the suggestion.

To get this to work we would need to either add a dependency on purrr or copy over the [`compat-purrr.R`](https://github.com/r-lib/rlang/blob/main/R/compat-purrr.R) file from rlang so we can use `imap()` and `discard()`. dplyr uses the same file for internal use of purrr functions.

@eutwt/@mgirlich Are you both good with adding `compat-purrr.R`? I think in the future it will be handy to have those functions available.

@eutwt - If you and @mgirlich both like the idea of using `compat-purrr.R` - would you be able to add the file and then make the suggested changes that @mgirlich mentioned [here](https://github.com/tidyverse/dtplyr/pull/324#issuecomment-1026641972)?

`imap_lgl()` doesn't currently exist in that file, but this is how it would be defined:

```r
imap_lgl <- function(.x, .y, .f, ...) {
  as.vector(imap(.x, .f, ...), "logical")
}
```

eutwt:
For ignoring trivial dots, I've added `&& .y %in% .data$vars)`, because this should still error:

```
lazy_dt(mtcars) %>% 
  mutate(zebra)
```

I like the idea of adding `compat-purrr.R`, but I think it's outside the scope of this PR.

markfairbanks:
Ah looks like we had `imap()` in the utils already - good catch.

> I like the idea of adding compat-purrr.R, but I think it's outside the scope of this PR.

I'm good with that. I'll go ahead and merge this one.

markfairbanks:
Also @eutwt - in the future now that you're an author feel free to directly make branches within `tidyverse/dtplyr`. It'll be a little bit easier for @mgirlich & I to make changes if necessary.

Issue #237, opened 2021-04-20
Translation to fifelse() and fcase() will throw error for S4 objects


torema-ed:
`fifelse` and `fcase` throws error when used on S4 objects as previously described (Rdatatable/data.table/issues/4131).

Would it be a good idea to not translate `if_else` and `case_when` when working with S4 columns or at least throw an error and suggest to run on a non-lazy_dt?

Reprex: 
``` r
# Loading packages --------------------------------------------------------

library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(dtplyr)
library(lubridate)
#> 
#> Attaching package: 'lubridate'
#> The following objects are masked from 'package:base':
#> 
#>     date, intersect, setdiff, union

# Creating test data ------------------------------------------------------

df <- data.frame(period_unit = c("days", "months", "weeks", "years"),
                 period_length = rep(1, times = 4))

# fcase output ------------------------------------------------------------

lazy_dt(df) %>%
  mutate(period = case_when(
    period_unit == "days" ~ days(period_length),
    period_unit == "months" ~ months(period_length),
    period_unit == "weeks" ~ weeks(period_length),
    period_unit == "years" ~ years(period_length)
  ))
#> Error in fcase(period_unit == "days", days(period_length), period_unit == : S4 class objects (except nanotime) are not supported. Please see https://github.com/Rdatatable/data.table/issues/4131.

# case_when output --------------------------------------------------------

df %>%
  mutate(period = case_when(
    period_unit == "days" ~ days(period_length),
    period_unit == "months" ~ months(period_length),
    period_unit == "weeks" ~ weeks(period_length),
    period_unit == "years" ~ years(period_length)
  ))
#>   period_unit period_length            period
#> 1        days             1       1d 0H 0M 0S
#> 2      months             1    1m 0d 0H 0M 0S
#> 3       weeks             1       7d 0H 0M 0S
#> 4       years             1 1y 0m 0d 0H 0M 0S
```

<sup>Created on 2021-04-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

markfairbanks:
[...]

PR #329, opened 2022-01-14
Do not translate namespaced calls

eutwt:
As shown below, tidyverse and base functions don't always behave the same as the similar data.table functions, so it would be nice to be able to avoid translation by using a namespaced function call.

On the other hand, this would make using dtplyr as a "drop in" speed-up for existing code slower if namespaced calls are used.

``` r
library(data.table)

dplyr::coalesce(factor('a'), factor('b'))
#> [1] a
#> Levels: a b
fcoalesce(factor('a'), factor('b'))
#> Error in fcoalesce(factor("a"), factor("b")): Item 2 is a factor but its levels are not identical to the first item's levels.

tidyr::replace_na(factor('a'), factor('b'))
#> Warning in `[<-.factor`(`*tmp*`, !is_complete(data), value =
#> structure(1L, .Label = "b", class = "factor")): invalid factor level, NA
#> generated
#> [1] a
#> Levels: a
fcoalesce(factor('a'), factor('b'))
#> Error in fcoalesce(factor("a"), factor("b")): Item 2 is a factor but its levels are not identical to the first item's levels.

dplyr::case_when(
  c(TRUE, FALSE) ~ factor('a'),
  c(TRUE, FALSE) ~ factor('b'))
#> Warning in `[<-.factor`(`*tmp*`, i, value = structure(1L, .Label = "b", class =
#> "factor")): invalid factor level, NA generated
#> [1] a    <NA>
#> Levels: a
fcase(
  c(TRUE, FALSE), factor('a'),
  c(TRUE, FALSE), factor('b'))
#> Error in fcase(c(TRUE, FALSE), factor("a"), c(TRUE, FALSE), factor("b")): Argument #2 and argument #4 are both factor but their levels are different.
          
base::ifelse(1, 'a', 'b')
#> [1] "a"
fifelse(1, 'a', 'b')
#> Error in fifelse(1, "a", "b"): Argument 'test' must be logical.

dplyr::if_else(c(TRUE, FALSE), factor('a'), factor('b'))
#> Warning in `[<-.factor`(`*tmp*`, i, value = structure(1L, .Label = "b", class =
#> "factor")): invalid factor level, NA generated
#> [1] a    <NA>
#> Levels: a
fifelse(c(TRUE, FALSE), factor('a'), factor('b'))
#> Error in fifelse(c(TRUE, FALSE), factor("a"), factor("b")): 'yes' and 'no' are both type factor but their levels are different.

dplyr::lag(1, 1:2)
#> Error: `n` must be a nonnegative integer scalar, not an integer vector of length 2.
shift(1, 1:2, type = 'lag')
#> [[1]]
#> [1] NA
#> 
#> [[2]]
#> [1] NA

dplyr::lead(1, 1:2)
#> Error: `n` must be a nonnegative integer scalar, not an integer vector of length 2.
shift(1, 1:2, type = 'lead')
#> [[1]]
#> [1] NA
#> 
#> [[2]]
#> [1] NA
```

<sup>Created on 2022-01-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/R/tidyeval.R
+++ b/R/tidyeval.R
@@ -104,10 +104,10 @@ dt_squash_call <- function(x, env, data, j = TRUE) {
     } else if (is_symbol(x[[2]], ".env")) {
       sym(paste0("..", var))
     }
-  } else if (is_call(x, c("coalesce", "replace_na"))) {
+  } else if (is_call(x, c("coalesce", "replace_na"), ns = '')) {
     args <- lapply(x[-1], dt_squash, env = env, data = data, j = j)
     call2("fcoalesce", !!!args)
-  } else if (is_call(x, "case_when")) {
+  } else if (is_call(x, "case_when", ns = '')) {
     # case_when(x ~ y) -> fcase(x, y)
     args <- unlist(lapply(x[-1], function(x) {
       list(
@@ -136,7 +136,7 @@ dt_squash_call <- function(x, env, data, j = TRUE) {
     x[[1]] <- sym("-")
     x[[2]] <- get_expr(x[[2]])
     x
-  } else if (is_call(x, c("if_else", "ifelse"))) {
+  } else if (is_call(x, c("if_else", "ifelse"), ns = '')) {
     if (is_call(x, "if_else")) {
       x <- unname(match.call(dplyr::if_else, x))
     } else {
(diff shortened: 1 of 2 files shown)


markfairbanks:
This idea was discussed in #237. I think @hadley [wants to take a different approach](https://github.com/tidyverse/dtplyr/issues/237#issuecomment-908340828). I'm not familiar with `dbplyr::local()`, so I'm not sure how that workaround would look.

eutwt:
Oh, I hadn't thought of using a wrapper function. I do think that's probably better, these cases seem rare.

PR #335, opened 2022-02-07
`across()` improvements

mgirlich:
~Fixes the third point in #287.~

* Refactor `dt_squash_if()` and `dt_squash_across()` to use `across_setup()`, very similar to `dplyr::across_setup()`.
* The naming is now more consistent with `dplyr`. When passing a list to `.fns` missing names are not guessed from the function but filled with a number.
* Characters cannot be used as function anymore (like in `dplyr`).
* Fixed some minor issues for `if_all()`.
* Increase test coverage for `if_all()`.
* Use `inject()` instead of deprecated `expr_interp()`


markfairbanks:
I like the simplification part of this PR - but before I do a code review I wanted to discuss the use of `...` with a purrr style lambda.

Basically I think we should error when `...` is used with a purrr style lambda (reasoning below).

In #287 @eutwt mentioned being able to do this:

``` r

df <- tibble(x = 1:3, y = 1:3)

df %>%
  mutate(
    across(c(x, y), ~ .x + .y, 1)
  )
#> # A tibble: 3 × 2
#>       x     y
#>   <dbl> <dbl>
#> 1     2     2
#> 2     3     3
#> 3     4     4
```

However `.y` here could just as easily be `..2`. And this can be extended to use more dots:

``` r

df <- tibble(x = 1:3, y = 1:3)

df %>%
  mutate(
    across(c(x, y), ~ .x + ..2 + ..3 + ..4, 1, 2, 3)
  )
#> # A tibble: 3 × 2
#>       x     y
#>   <dbl> <dbl>
#> 1     7     7
#> 2     8     8
#> 3     9     9
```

@mgirlich maybe you have a good idea to deal with situations like this with an arbitrary number of dots. But instead of supporting `.y` (or the case of only one dot with a purrr style lambda), I would rather error if dots are used.

[In the `across()` docs for dplyr 1.0.8 it mentions the use of dots is discouraged](https://github.com/tidyverse/dplyr/blob/main/R/across.R#L34). So I think it makes sense to have them error in this edge case in `dtplyr`.

eutwt:
Although I find the ability to use `...` in `across` useful, I think I agree dtplyr probably should not support it given that it's [likely to be deprecated in dplyr 1.1.0](https://github.com/tidyverse/dplyr/issues/6073). Apologies for contradicting my own issue / not keeping that issue current.

mgirlich:
Fine for me to remove to support for dots, especially for purrr style lambdas.
I didn't even know you could use `..2`, `..3`, etc for purrr style lambdas and to me your example just looks... terrible.

markfairbanks:
I think we should definitely error in the case where dots are used with a purrr-style lambda. It doesn't currently work anyway so it won't break backwards compatibility at all.

However deprecating dots usage in general should be done with a deprecation lifecycle (if at all) since it currently works in dtplyr and also works in dplyr. I think a case like this should still work (without warnings) until dplyr makes the deprecation first.

```r
df <- data.table(x = 1:3, y = 1:3)

df %>%
  summarize(
    across(c(x, y), mean, na.rm = TRUE)
  )
```

R/tidyeval-across.R
```diff
@@ -5,90 +5,61 @@ capture_across <- function(data, x, j = TRUE) {
...
+    fns <- list(`1` = across_fun(funs, env, data, j = j, dots = dots))
+    names_spec <- names_spec %||% "{.col}"
   } else if (is_call(funs, "~")) {
-    set_names(list(across_fun(funs, env, data, j = j)), expr_name(f_rhs(funs)))
+    fns <- list(`1` = across_fun(funs, env, data, j = j, dots = dots))
+    names_spec <- names_spec %||% "{.col}"
   } else if (is_call(funs, "list")) {
-    args <- rlang::exprs_auto_name(funs[-1])
-    lapply(args, across_fun, env, data, j = j)
+    args <- rlang::call_args(funs)
```
markfairbanks, line 34: Can drop `rlang::` here

PR #336, opened 2022-02-07
Add compat purrr

mgirlich:
Add compat purrr and use `map_*()` instead of `vapply()`


markfairbanks:
Looks good to me - feel free to merge whenever you're ready.

PR #338, opened 2022-02-08
skip test relying on rlang>=1.0.0 behavior

MichaelChirico:


markfairbanks:
dtplyr now depends on rlang >= 1.0.0, so there shouldn't be a need to skip this test anymore

PR #343, opened 2022-02-11
Implement `unite()`

### LINKED ISSUE #301: Implement `unite()`
Hello, great work on the `dtplyr` package, its  a really nice addition to the `tidyverse`. I was wondering if there will be more methods that extend some of `tidyr` generics. This is more or less an extended feature request  to #266.

Most people who've adapted to using `dplyr` have likely relied on a few `tidyr` functions.

I think many people would benefit from the addition of:

* nest.dtplyr_step
* unnest.dtplyr_step
* fill.dtplyr_step
* pivot_longer.dtplyr_step (to pair with the existing `pivot_wider.dtplyr_step`)
* separate.dtplyr_step
* unite.dtplyr_step

There already exists a package out there that attempts to address these functions named `tidyfast`. Only downside is that these functions would force a lazy_dt to execute its steps as I believe the first step in `tidyfast` functions is to coerce the object with `as.data.table()`

markfairbanks:
Closes #301

Note: This errors on `na.rm = TRUE` as there isn't a clean way to make the translation for a `lazy_dt` (at least from what I could tell).


R/unite.R
```diff
@@ -0,0 +1,56 @@
...
+    sim_data <- simulate_vars(data)
+    .cols <- names(tidyselect::eval_select(expr(c(!!!dots)), sim_data))
+  }
+
+  out <- mutate(ungroup(data), !!.col := paste(!!!syms(.cols), sep = sep))
+
+  remove <- is_true(remove)
+  if (remove) {
+    .drop_cols <- setdiff(.cols, .col)
+    out <- select(out, -tidyselect::all_of(.drop_cols))
```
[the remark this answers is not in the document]
eutwt: I guess this is true of select generally. Maybe better to address this in select itself, rather than the specific case in this PR as I was thinking earlier.

R/unite.R
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  .col <- as_name(enquo(col))
+
+  dots <- enquos(...)
+  if (length(dots) == 0) {
+    .cols <- data$vars
+  } else {
+    sim_data <- simulate_vars(data)
+    .cols <- names(tidyselect::eval_select(expr(c(!!!dots)), sim_data))
```
mgirlich, line 33: This should probably use `allow_rename = FALSE`, e.g. something like

```r
locs <- tidyselect::eval_select(expr(c(x, y)), sim_data, allow_rename = FALSE)
.cols <- data$vars[locs]
```

Otherwise, to prevent this error (which also happens for normal dataframes)

```r
df %>% unite("z", c(a = x, y), remove = FALSE)
#> Error in paste(a, y, sep = "_") : object 'a' not found
```

Note that the `allow_rename = FALSE` does not really do anything here, using the locations is the relevant part here. But I think `allow_rename = FALSE` communicates the intent better.

R/unite.R
```diff
@@ -0,0 +1,56 @@
...
+    .drop_cols <- setdiff(.cols, .col)
+    out <- select(out, -tidyselect::all_of(.drop_cols))
+  }
+
+  group_vars <- data$groups
+  if (remove && any(.cols %in% group_vars)) {
+    out
+  } else {
+    group_by(out, !!!syms(group_vars))
+  }
```
mgirlich, line 49: `unite()` keeps the groups where possible:

```r
df <- lazy_dt(expand_grid(x = c("a", NA), y = c("b", NA)))
df %>% group_by(x, y) %>% unite("z", x)
df %>% collect() %>% group_by(x, y) %>% unite("z", x)
```

So, you could replace it by something like

```r
new_groups <- setdiff(group_vars, .cols)
group_by(out, !!!syms(new_groups))
```

and add a test that the group is kept.

R/unite.R
```diff
@@ -0,0 +1,56 @@
...
+
+  dots <- enquos(...)
+  if (length(dots) == 0) {
+    .cols <- data$vars
+  } else {
+    sim_data <- simulate_vars(data)
+    .cols <- names(tidyselect::eval_select(expr(c(!!!dots)), sim_data))
+  }
+
+  out <- mutate(ungroup(data), !!.col := paste(!!!syms(.cols), sep = sep))
```
mgirlich, line 36: The order of the variables is not correct

```r
df <- lazy_dt(expand_grid(x = c("a", NA), y = c("b", NA)))
df %>% collect() %>% unite("z", x) %>% colnames()
#> "z" "y"
df %>% unite("z", x) %>% .$vars
#> "y" "z"
```

R/unite.R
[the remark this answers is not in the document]
markfairbanks: > Maybe better to address this in select itself, rather than the specific case in this PR as I was thinking earlier.

Sounds good. I'll leave this part as is.

markfairbanks:
@mgirlich @eutwt Do either of you know how to fix this issue? The first test fails because the translation is "too long" and building it from scratch using `expr()` causes weird spacing issues. The translation is correct. Here's a reprex:

``` r
devtools::load_all(".")

df <- lazy_dt(data.table(x = "a", y = "b"), "DT")
step <- unite(df, "z", x:y)

show_query(step)
#> setcolorder(copy(DT)[, `:=`(z = paste(x, y, sep = "_"))], c("z", 
#> "x", "y"))[, .(z)]
expr(setcolorder(copy(DT)[, `:=`(z = paste(x, y, sep = "_"))], c("z", "x", "y"))[, .(z)])
#> setcolorder(copy(DT)[, `:=`(z = paste(x, y, sep = "_"))], c("z", 
#>     "x", "y"))[, .(z)]
```

markfairbanks:
> For me, if you put !! in front of c("z", "x", "y") in the expression you create, the test passes.

This worked, thanks @eutwt

markfairbanks:
@mgirlich All set for review.

R/unite.R
```diff
@@ -0,0 +1,61 @@
...
+  dots <- enquos(...)
+  if (length(dots) == 0) {
+    .cols <- data$vars
+  } else {
+    sim_data <- simulate_vars(data)
+    locs <- tidyselect::eval_select(expr(c(!!!dots)), sim_data, allow_rename = FALSE)
+    .cols <- data$vars[locs]
+  }
+
+  out <- mutate(ungroup(data), !!.col := paste(!!!syms(.cols), sep = sep), .before = !!sym(.cols[[1]]))
```
mgirlich, line 37: Does not work correctly if cols are specified in a different order

```r
df %>% collect() %>% unite("z", c(y, x), remove = FALSE) %>% colnames()
#> "z" "x" "y"

df %>% unite("z", c(y, x), remove = FALSE) %>% .$vars
#> "x" "z" "y"
```

This should use something like `min(locs)`.

You could also produce a shorter query by using `relocate()` explicitly

```r
out <- mutate(ungroup(data), !!.col := paste(!!!syms(.cols), sep = sep))

  remove <- is_true(remove)
  if (remove) {
    .drop_cols <- setdiff(.cols, .col)
    out <- select(out, -tidyselect::all_of(.drop_cols))
  }
  out <- relocate(out, !!.col, .before = min(locs))
```

mgirlich (approved):

PR #354, opened 2022-04-22
Add `.keep` to `mutate`

### LINKED ISSUE #340: Add `.keep` to `mutate()`
``` r

df <- tibble(x = 1:3, y = 1:3)

df %>%
  mutate(double_x = x * 2, .keep = "used")
#> # A tibble: 3 × 2
#>       x double_x
#>   <int>    <dbl>
#> 1     1        2
#> 2     2        4
#> 3     3        6
```

markfairbanks:
Closes #340 

A couple other small changes that didn't seem worth their own PR:
* Bumped `vctrs` dependency because a couple snapshots were using the new version. Bumped to `v0.4.1`
* There's a small doc update to `group_by` that keeps popping up every time I run `document()`

### DIFF
--- a/R/step-mutate.R
+++ b/R/step-mutate.R
@@ -89,32 +89,33 @@ mutate_with_braces <- function(mutate_vars) {
 mutate.dtplyr_step <- function(.data, ...,
                                .keep = c("all", "used", "unused", "none"),
                                .before = NULL, .after = NULL) {
-  dots <- capture_new_vars(.data, ...)
-  trivial_dot <- imap(dots, ~ is_symbol(.x) && sym(.y) == .x && .y %in% .data$vars)
-  dots <- dots[!as.vector(trivial_dot, "logical")]
+  all_dots <- capture_new_vars(.data, ...)
+  trivial_dot <- imap(all_dots, ~ is_symbol(.x) && sym(.y) == .x && .y %in% .data$vars)
+  dots <- all_dots[!as.vector(trivial_dot, "logical")]
   dots_list <- process_new_vars(.data, dots)
   dots <- dots_list$dots
+
   if (is_null(dots) || is_empty(dots)) {
-    return(.data)
-  }
+    out <- .data
+  } else {
+    out <- step_mutate(.data, dots, dots_list$use_braces)
 
-  out <- step_mutate(.data, dots, dots_list$use_braces)
+    .before <- enquo(.before)
+    .after <- enquo(.after)
+    if (!quo_is_null(.before) || !quo_is_null(.after)) {
+      # Only change the order of new columns
+      new <- setdiff(names(dots), .data$vars)
+      out <- relocate(out, !!new, .before = !!.before, .after = !!.after)
+    }
 
-  .before <- enquo(.before)
-  .after <- enquo(.after)
-  if (!quo_is_null(.before) || !quo_is_null(.after)) {
-    # Only change the order of new columns
-    new <- setdiff(names(dots), .data$vars)
-    out <- relocate(out, !!new, .before = !!.before, .after = !!.after)
-  }
-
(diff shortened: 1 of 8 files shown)


eutwt:
This will be nice to have. Will look in more detail later, but one thing I notice is that the edge cases of empty dots and trivial dots need special care, e.g. 

```r
lazy_dt(mtcars) %>% mutate(mpg, .keep = 'used')  
# vs
mtcars %>% mutate(mpg, .keep = 'used')  

lazy_dt(mtcars) %>% mutate(.keep = 'used')  
# vs
mtcars %>% mutate(.keep = 'used')  
```

R/step-mutate.R
```diff
@@ -56,6 +56,19 @@ mutate_with_braces <- function(mutate_vars) {
 #' @param ... <[data-masking][dplyr::dplyr_data_masking]> Name-value pairs.
 #'   The name gives the name of the column in the output, and the value should
 #'   evaluate to a vector.
+#' @param .keep `r lifecycle::badge("experimental")`
```
eutwt, line 59: Might need some note here, maybe just at the end, explaining that this is based on the symbols present in the generated `data.table` code. Stuff (rarely used) like `get()` and `eval()` will result in different output than dplyr. For example, dplyr returns a three-column tibble with `mutate(mtcars, y = eval(parse(text = "wt + hp")), .keep = "used")`

eutwt, line 59: Might need some note here, maybe just at the end, explaining that this is based on the symbols present in the generated `data.table` code. Stuff (rarely used) like `get()` and `eval()` will result in different output than dplyr. For example, dplyr returns a three-column tibble with `mutate(mtcars, y = eval(parse(text = "wt + hp")), .keep = "used")`

markfairbanks:
@eutwt @mgirlich - any idea how to deal with this snapshot error causing a failed workflow run? The snapshot was updated in [1903d3e](https://github.com/tidyverse/dtplyr/pull/354/commits/1903d3eef2ec71002aed8ad2f1fdcff15fc9d992) but for some reason it fails on `macOS-latest`. It's fine locally and in all other checks. Ironically enough I'm on the latest macOS.

Other than this strange error this is all ready for review.

eutwt:
I think it's because the macOS-latest check uses vctrs 0.4.0 which has the "past the end" wording, and the others are using vctrs 0.4.1 which reverted that change to the wording

markfairbanks:
Yep that fixed it - thanks @eutwt

eutwt:
I think this is good to merge, except for style considerations, see below


I notice that [dbplyr](https://github.com/tidyverse/dplyr/blob/1cff4b9ce7f404955d8a5764b5708eee5972e0e3/R/mutate.R#L196) is using the exact same setup as [dplyr](https://github.com/tidyverse/dbplyr/blob/e371ffd47c86eb1c46725d3fbcff1a70d06d3376/R/verb-mutate.R#L58) for the `.keep` argument. It's possible to rewrite the code in this PR to be the same as those ([example](https://github.com/tidyverse/dtplyr/commit/367444052abde40587f4b24cc3739e1f9dba4875)), and I think that's worth considering just to keep close to dbplyr. Staying close to dbplyr is especially attractive since dbplyr appears far ahead of dtplyr in terms of combing multiple steps into a single query. Take for example `lazy_dt(mtcars) %>% mutate(a = mpg, d = 2, d = NULL, .keep = 'unused') ` which is 3 queries in dtplyr and one in dbplyr. At some point it would be nice to move dtplyr in line with the dbplyr layer approach, which would be a little easier if our code was similar.

On the other hand your current code appears more compact, and it probably doesn't make too much difference either way. But, thought I'd mention the consideration

eutwt (approved):

PR #363, opened 2022-06-21
Error when `where()` is used in `across()`

### LINKED ISSUE #271: summarize(across(where)) fails when using a predicate and acts unexpectedly with predicate negation
I have been trying to summarize across numeric columns using dtplyr and `summarise(across(where(is.numeric)))` and got the error `can't rename variables in this context`. It took me quite a while to figure out the reason for the error is the usage of a predicate within `summarize` is not possible with dtplyr.
Is there any substitute for using `is.numeric` besides for naming all columns?
Thank you

The following code 1) shows error when using predicate, 2) show how the negation does not fail and applies given function to all variables (numeric and non-numeric)
```r
library(ggplot2)
library(data.table)
library(dtplyr)

# fails
diamonds %>% lazy_dt() %>% group_by(color) %>% summarise(across(!where(is.numeric), first))

# applies function to all variables
diamonds %>% lazy_dt() %>% group_by(color) %>% summarise(across(!where(is.numeric), first))

# unlike the equivalent statement using a data frame which only returns the relevant variables
diamonds %>% group_by(color) %>% summarise(across(!where(is.numeric), first))
```

markfairbanks:
Closes #271

The biggest thing this prevents is the "silent failure" cases.
``` r
library(dtplyr)

df <- tibble(x = 1:3, y = c("a", "b", "c"), z = c("a", "a", "b"))

lazy_dt(df) %>%
  group_by(z) %>%
  summarize(across(!where(is.numeric), first))
#> Source: local data table [2 x 3]
#> Call:   `_DT1`[, .(x = first(x), y = first(y)), keyby = .(z)]
#> 
#>   z         x y    
#>   <chr> <int> <chr>
#> 1 a         1 a    
#> 2 b         3 c    
#> 
```

Note: This can be updated later if https://github.com/r-lib/tidyselect/issues/226 is implemented.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # dtplyr (development version)
 
+* An error now occurs when `where()` is used in `across()` (#271)
+
 * Custom functions can pass a quosure to `col` arg in `separate()` (#359)
 
 * `glue::glue()` and `stringr::str_glue()` now work inside `mutate()`/`transmute()` without needing
--- a/R/tidyeval-across.R
+++ b/R/tidyeval-across.R
@@ -156,6 +156,7 @@ across_glue_mask <- function(.col, .fn, .caller_env) {
 }
 
 across_uses_where <- function(x) {
+  x <- quo_squash(x)
   if (is_call(x, "where")) {
     TRUE
   } else {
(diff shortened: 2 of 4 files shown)


eutwt:
Should we just implement a temporary `eval_select` wrapper in dtplyr that disallows predicates until that happens? That way this would be fixed for other cases too. In dplyr you can use `where` in select, relocate, rename_with, mutate's .before and .after, and it looks like all of those are silent failures in dtplyr.

markfairbanks:
Yeah I like that option better - I'll close this PR and we can do that in a different one.

markfairbanks:
FYI @eutwt I opened an issue for adding the helper

PR #365, opened 2022-06-23
`dt_squash` anonymous functions

### LINKED ISSUE #362: Better handling of anonymous functions by internal `capture_dots()`
``` r
library(dplyr, w = FALSE)
library(dtplyr)
library(data.table, w = FALSE)

df <- data.table(x = list(1, 1, 1))

df %>%
  mutate(z = sapply(x, function(.x) .x + n()))
#> Error in `n()`:
#> ! Must be used inside dplyr verbs.
```

markfairbanks:
Closes #362

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # dtplyr (development version)
 
+* Anonymous functions are translated correctly in `mutate`/`summarize` (#362)
+
 * Custom functions can pass a quosure to `col` arg in `separate()` (#359)
 
 * `glue::glue()` and `stringr::str_glue()` now work inside `mutate()`/`transmute()` without needing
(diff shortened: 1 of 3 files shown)


eutwt (approved):
LGTM

eutwt:
This fixes the issue outside of across, but make me realize I didn't handle #288 right, because anonymous functions in `across` don't get dplyr verbs translated

```r
lazy_dt(tibble(y = 1)) %>% 
  mutate(across(y, function(x) x + n()))
```

markfairbanks:
Ah for some reason I assumed that `across()` would be using the same `dt_squash()` path but looks like it's different.

Might be easier to merge this one and deal with the `across()` issue separately. That work for you @eutwt?

eutwt:
Yeah I think that makes sense

PR #366, opened 2022-06-23
Properly copy with chained operations

### LINKED ISSUE #210: Input data.table can be modified despite having immutable=TRUE in lazy_dt()
Despite specifying `immutable=TRUE` in `lazy_dt`, if I have a `mutate` followed by a `summarize` followed by a `mutate`, the first `mutate` will get applied to the input data.table object, when it should not be modified. I can't seem to trigger it unless I have a `mutate` after the `summarize`.

``` r
library(data.table)
library(dtplyr)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:data.table':
#> 
#>     between, first, last
#> 
#>     filter, lag
#> 

d1= data.table(x=c(1,1,1,2,2,2), y=c(1,2,3,4,5,6))

d2 = lazy_dt(d1, immutable = TRUE) %>%
  mutate(
    y=-y,
    z = x+y
    ) %>%
  summarize(
    mean_y = mean(y),
    mean_z = mean(z)
    ) %>%
  mutate(
    sum_mean = mean_y + mean_z
  ) %>%
  as.data.table()


d1
#>    x  y  z
#> 1: 1 -1  0
#> 2: 1 -2 -1
#> 3: 1 -3 -2
#> 4: 2 -4 -2
#> 5: 2 -5 -3
#> 6: 2 -6 -4
```

<details>

Here's my `sessionInfo()`:
``` r
R version 4.0.3 (2020-10-10)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 19042)

Matrix products: default

locale:
[1] LC_COLLATE=English_Canada.1252  LC_CTYPE=English_Canada.1252    LC_MONETARY=English_Canada.1252 LC_NUMERIC=C                   
[5] LC_TIME=English_Canada.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] dplyr_1.0.3       dtplyr_1.1.0      data.table_1.13.6

loaded via a namespace (and not attached):
 [1] assertthat_0.2.1 crayon_1.4.0     R6_2.5.0         DBI_1.1.1        lifecycle_0.2.0  magrittr_2.0.1   pillar_1.4.7     rlang_0.4.10    
 [9] vctrs_0.3.6      generics_0.1.0   ellipsis_0.3.1   tools_4.0.3      glue_1.4.2       purrr_0.3.4      compiler_4.0.3   pkgconfig_2.0.3 
[17] tidyselect_1.1.0 tibble_3.0.5 
```

</details>

markfairbanks:
Closes #210 

Basic idea - once `needs_copy` is made `TRUE` it is kept as `TRUE` for the rest of the operations.


eutwt:
Looks good to merge. 

Poking around I did find another case where strictly speaking the original object is modified, because `filter` can kick off data.table's auto-indexing. I think we'd want to keep that though, because the alternative would be to copy for every filter, which doesn't seem like a good tradeoff. I don't think auto-indexing on the original data.table has any downside I can see, except that the index takes a little memory. 

``` r
devtools::load_all('~/Documents/GitHub/dtp')
#> ℹ Loading dtplyr
#> Warning: package 'dplyr' was built under R version 4.1.2

dt <- data.table(a = sample(1e7))
attr(dt, "index")
#> NULL

lazy_dt(dt) %>% filter(a == 1)
#> Call:   `_DT1`[a == 1]
#> 
#>       a
#>   <int>
#> 1     1
#> 
attr(dt, "index") %>% attr("__a") %>% str
#>  int [1:10000000] 5568870 634911 7756538 1471857 7904572 5964142 1696470 8336329 5935559 2557143 ...
attr(dt, "index") %>% object.size() %>% print(units = 'Mb')
#> 38.1 Mb
```

<sup>Created on 2022-07-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

eutwt (approved):

mgirlich (approved):

PR #369, opened 2022-06-25
Implement internal `dtplyr_tidyselect()`

### LINKED ISSUE #271: summarize(across(where)) fails when using a predicate and acts unexpectedly with predicate negation

### LINKED ISSUE #368: Implement internal `eval_select()` helper that errors with `where()`
See #271 

See #363 & https://github.com/tidyverse/dtplyr/pull/363#issuecomment-1165886564

markfairbanks:
Closes #271 
Closes #368

### DIFF
--- a/R/step-subset-select.R
+++ b/R/step-subset-select.R
@@ -17,8 +17,7 @@
 #' dt %>% select(ends_with("2"))
 #' dt %>% select(z1 = x1, z2 = x2)
 select.dtplyr_step <- function(.data, ...) {
-  sim_data <- simulate_vars(.data)
-  locs <- tidyselect::eval_select(expr(c(...)), sim_data)
+  locs <- dtplyr_tidyselect(.data, ...)
   locs <- ensure_group_vars(locs, .data$vars, .data$groups)
 
   vars <- set_names(.data$vars[locs], names(locs))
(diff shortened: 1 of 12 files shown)


R/step-subset-select.R
```diff
@@ -47,6 +46,34 @@ select.data.table <- function(.data, ...) {
   select(.data, ...)
 }
 
+dtplyr_tidyselect <- function(.data, ...,
+                              .env = caller_env(),
+                              .allow_rename = TRUE,
+                              .drop_groups = FALSE) {
+  dots <- enquos(...)
+  if (selection_uses_where(dots)) {
+    abort("The use of `where()` is not supported by dtplyr.")
```
eutwt, line 55: I think we want `abort(..., call = .env)` here so we get the name of the top-level function in the error e.g. 

```r
lazy_dt(tibble(y = 1)) %>% 
  select(where(y))
#> Error in `select()`:
#> ! The use of `where()` is not supported by dtplyr.
```

instead of

```r
lazy_dt(tibble(y = 1)) %>% 
  select(where(y))
#> Error in `dtplyr_tidyselect()` at dtp/R/step-subset-select.R:20:2:
#> ! The use of `where()` is not supported by dtplyr.
```

matching dplyr

```r
tibble(y = 1) %>% 
  select(where(fakefun))
#> Error in `select()`:
#> ! object 'fakefun' not found
```

eutwt: Hmm, looks like `.env` isn't passed around the right way for this to work in `mutate(across`, because the change I suggested above results in a blank first line in the code below. Not sure if it's worth trying to fix our errors in this PR. It would be good to get `where()` to stop silently failing and the error is still unambiguous, this is more of a minor dplyr-consistency thing.

```r
mutate(dt, across(!where(is.character), ~ .x + 1))
#> Error:
#> ! The use of `where()` is not supported by dtplyr.
```

markfairbanks: I was able to fix this by passing `.call = call2(fn)` in `across_setup()`. So whether using `across()` or `if_any()`/`if_all()` we now have the proper error message.

eutwt (approved):
LGTM

Issue #377, opened 2022-07-19
`slice()` unnecessarily evaluates `...` twice for each group


eutwt:
`slice()` could be made faster when the arguments in `...` are expensive to compute by evaluating `...` only once per group. In the example below where the argument to slice takes one second to compute and there are two groups, the code takes ~4 seconds to run instead of ~2.

``` r
library(dtplyr)
library(dplyr, warn.conflicts = FALSE)
#> Warning: package 'dplyr' was built under R version 4.1.2
library(data.table, warn.conflicts = FALSE)

dt <- data.table(id = c(1, 2))

return_1 <- function() {Sys.sleep(1); 1}

library(tictoc)
tic()
dt %>% 
  lazy_dt() %>% 
  group_by(id) %>% 
  slice(return_1())
#> Source: local data table [2 x 1]
#> Groups: id
#> Call:   `_DT1`[`_DT1`[, .I[return_1()[between(return_1(), -.N, .N)]], 
#>     by = .(id)]$V1]
#> 
#>      id
#>   <dbl>
#> 1     1
#> 2     2
#> 
#> # Use as.data.table()/as.data.frame()/as_tibble() to access results
toc()
#> 4.074 sec elapsed

tic()
dt[dt[,
  {
    slice_ind <- return_1()
    .I[slice_ind[between(slice_ind, -.N, .N)]]
  },
  by = .(id)
]$V1]
#>    id
#> 1:  1
#> 2:  2
toc()
#> 2.014 sec elapsed
```

<sup>Created on 2022-07-19 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

markfairbanks:
I sort of still like the idea of doing this one 🤷‍♂️

`data.table` seems to be pretty slow getting this release out. If we release before they do it'd be nice to have.

PR #378, opened 2022-07-21
Use intermediate variable in `slice()`

markfairbanks:
See #377

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -41,6 +41,8 @@
 
 * `arrange()` now utilizes `setorder()` when possible for improved performance (#364)
 
+* `slice()` uses an intermediate variable to reduce computation time of row selection (#377)
+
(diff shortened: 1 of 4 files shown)


markfairbanks:
Good for review.

Also does anyone have an idea what this check failure is caused by? 🤷‍♂️

eutwt (approved):

eutwt:
My guess is that since the output of `expr({...})` has a "srcref" attribute, covr treats it as a separate source file to check. You can either use `call2` instead (whose output doesn't have that attribute) or use `i <- rlang::zap_srcref(i)` after to get rid of the attribute.

mgirlich (approved):

markfairbanks:
Thanks for figuring out that test @eutwt 😄

PR #379, opened 2022-07-26
Show user-level functions in tidyselect errors.

eutwt:
Following the versions used for rlang and tidyselect in https://github.com/tidyverse/dplyr/commit/eaac641f6c20e050706b46db2558f79a92504d58

### DIFF
--- a/R/step-subset-select.R
+++ b/R/step-subset-select.R
@@ -61,7 +61,10 @@ dtplyr_tidyselect <- function(.data, ...,
     abort("The use of `where()` is not supported by dtplyr.", call = .call)
   }
   sim_data <- simulate_vars(.data, drop_groups = .drop_groups)
-  tidyselect::eval_select(expr(c(!!!dots)), sim_data, allow_rename = .allow_rename)
+  tidyselect::eval_select(expr(c(!!!dots)), sim_data,
+    allow_rename = .allow_rename,
+    error_call = .call
+  )
 }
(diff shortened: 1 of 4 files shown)


R/step-subset-select.R
```diff
@@ -61,7 +61,10 @@ dtplyr_tidyselect <- function(.data, ...,
     abort("The use of `where()` is not supported by dtplyr.", call = .call)
   }
   sim_data <- simulate_vars(.data, drop_groups = .drop_groups)
-  tidyselect::eval_select(expr(c(!!!dots)), sim_data, allow_rename = .allow_rename)
+  tidyselect::eval_select(expr(c(!!!dots)), sim_data,
```
markfairbanks, line 64: Minor formatting adjustment - can this be changed to one of these 2 options?

Up to you which one. I personally think Option 2 looks a bit odd and doesn't work well with RStudio's auto-indent but that pattern is used elsewhere in dtplyr.

``` r
# Option 1
tidyselect::eval_select(
  expr(c(!!!dots)),
  sim_data,
  allow_rename = .allow_rename,
  error_call = .call
)

# Option 2
tidyselect::eval_select(expr(c(!!!dots)),
  sim_data,
  allow_rename = .allow_rename,
  error_call = .call
)
```

Other than that this looks good to merge.

markfairbanks (approved):

PR #387, opened 2022-08-16
Inject syms that start with "." but aren't data.table syms

### LINKED ISSUE #386: Function calls in functions can not access to environment variables in parent functions
@markfairbanks Still fails in such circumstances (seemingly, function calls in functions can not access environment variables, either):

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(dtplyr)
f <- function(x, y, .v = "x") {
  inner_join(
    mutate(x, a = .v),
    y,
    by = character()
  )
}
x <- data.frame(a = "y")
y <- data.frame(b = "x")
f(x, y)
#>   a b
#> 1 x x
lazy_dt(x) |> 
  f(y)
#> Error in .checkTypos(e, names_x): Object '.v' not found. Perhaps you intended .cross_join_col
```

<sup>Created on 2022-08-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

_Originally posted by @psychelzh in https://github.com/tidyverse/dtplyr/issues/317#issuecomment-1214932833_

eutwt:
closes #386

``` r
library(dplyr, w = FALSE)
library(dtplyr, w = FALSE)

f <- function(x, .v) {
  mutate(x, a = .v)
}

x <- data.frame(a = "y")

lazy_dt(x) |>
  f(.v = "injected")
#> Call:   copy(`_DT1`)[, `:=`(a = "injected")]
#> 
#>   a       
#>   <chr>   
#> 1 injected
#> 
```

<sup>Created on 2022-08-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

### DIFF
--- a/R/tidyeval.R
+++ b/R/tidyeval.R
@@ -18,6 +18,7 @@ dt_funs <- c(
   "fcase", "fcoalesce", "fifelse", "fintersect", "frank", "frankv", "fsetdiff", "funion",
   "setcolorder", "setnames", "setorder", "shift", "tstrsplit", "uniqueN"
 )
+dt_symbols <- c(".SD", ".BY", ".N", ".I", ".GRP", ".NGRP")
 add_dt_wrappers <- function(env) {
   env_bind(env, !!!env_get_list(ns_env("data.table"), dt_funs))
 }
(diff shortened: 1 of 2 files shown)


eutwt:
Another option would be to just delete lines 80-82. I think the only situation it affects is a user-supplied data.table pronoun symbol matching an environment variable. So if a user calls `mutate(lazydt, y = .N)` and they have e.g. `.N <- 4` as an environment variable, they get a query like `dt[, y := .N]` instead of `dt[, y := 4]`.

This seems pretty rare, and when it does happen we're creating behavior that diverges from dplyr. But, maybe people are relying on the existing bahavior?

markfairbanks:
I think you're right - deleting 80-82 would probably be fine. But I think the `dt_symbols` approach is the right one to use since it preserves backwards compatibility.

markfairbanks (approved):

PR #396, opened 2022-10-25
Translate `min_rank`, `dense_rank`, `percent_rank`, & `cume_dist`

markfairbanks:


eutwt (approved):
Looks good. Would be nice to have a single-argument assertion for these too, like with `desc()` https://github.com/tidyverse/dtplyr/blob/dae58ece18c122a7040cb4373ba2bc166fe74a8d/R/tidyeval.R#L154-L157

PR #404, opened 2022-12-14
Use snapshots for errors that dplyr owns now

DavisVaughan:
This PR makes your package compatible with the next version of dplyr:

`slice_*()` now handles some of the checks in the generic itself, and the error messages it throws are different from what you had before. I've made these snapshot error tests since dplyr really owns the error message now, and we don't want this to fail on CRAN if dplyr ever changes it.

We plan to submit dplyr 1.1.0 on January 27th.

This should be compatible with both dev and CRAN dplyr. It would help us out if you could go ahead and send a patch version of your package to CRAN ahead of time! Thanks!

### DIFF
--- a/tests/testthat/_snaps/step-subset-slice.md
+++ b/tests/testthat/_snaps/step-subset-slice.md
@@ -4,32 +4,42 @@
       slice_head(dt, 5)
     Condition
       Error in `slice_head()`:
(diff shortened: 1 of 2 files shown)


markfairbanks:
Sounds good, thanks @DavisVaughan

PR #416, opened 2023-02-21
Remove out of date sections

markfairbanks:
Removed out of date v1.0.0 comment.

Also removed a section from "Why is dtplyr slower than data.table?".

```
* Some data.table expressions have no direct dplyr equivalent. For example,
  there's no way to express cross- or rolling-joins with dplyr.
```

I couldn't think of a good example here, let me know if I should add it back in.

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -25,8 +25,6 @@ knitr::opts_chunk$set(
 
 dtplyr provides a [data.table](http://r-datatable.com/) backend for dplyr. The goal of dtplyr is to allow you to write dplyr code that is automatically translated to the equivalent, but usually much faster, data.table code.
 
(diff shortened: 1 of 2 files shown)


README.Rmd
```diff
@@ -83,17 +81,14 @@ mtcars2 %>%
...
-There are three primary reasons that dtplyr will always be somewhat slower than data.table:
+There are two primary reasons that dtplyr will always be somewhat slower than data.table:
 
 * Each dplyr verb must do some work to convert dplyr syntax to data.table 
   syntax. This takes time proportional to the complexity of the input code, 
   not the input _data_, so should be a negligible overhead for large datasets.
   [Initial benchmarks][benchmark] suggest that the overhead should be under 
   1ms per dplyr call.
 
-* Some data.table expressions have no direct dplyr equivalent. For example,
```
hadley, line 94: I'm sure there still are things, but they're probably pretty small/minor now.

hadley (approved):

PR #422, opened 2023-03-21
Upkeep

hadley:


hadley:
@markfairbanks @mgirlich @eutwt this fixes a failing check on R-devel; I'll submit a patch release to CRAN later today.

PR #428, opened 2023-03-28
Allow namespaced calls in translated functions

### LINKED ISSUE #427: Issue when namespacing desc function in an arrange call
I have an arrange call with a desc call nested inside it, which I am trying to pass through lazy_dt().

The actual implementation is nested in a function in a package, hence the namespacing. However, the issue still presents itself outside of that context when the namespacing is introduced, as per the reprex below. 

I can drop the namespacing of the function in the package code, and add @importFrom dplyr desc to the roxygen header, which seems to get me around the issue for now, but seemed worth raising in case there are other implications.

Tested on R 4.2.1 and R 4.2.2, the code works fine with dplyr 1.0.10 and dtplyr 1.2.2 (versions I had before recently upgrading) and fails in dplyr 1.1.1 and dtplyr 1.3.1

``` r
library(dplyr)
library(dtplyr)

mtcars |> 
  lazy_dt() |> 
  arrange(desc(vs), cyl)
#> Source: local data table [32 x 11]
#> Call:   `_DT1`[order(-vs, cyl)]
#> 
#>     mpg   cyl  disp    hp  drat    wt  qsec    vs    am  gear  carb
#>   <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
#> 1  22.8     4 108      93  3.85  2.32  18.6     1     1     4     1
#> 2  24.4     4 147.     62  3.69  3.19  20       1     0     4     2
#> 3  22.8     4 141.     95  3.92  3.15  22.9     1     0     4     2
#> 4  32.4     4  78.7    66  4.08  2.2   19.5     1     1     4     1
#> 5  30.4     4  75.7    52  4.93  1.62  18.5     1     1     4     2
#> 6  33.9     4  71.1    65  4.22  1.84  19.9     1     1     4     1
#> # … with 26 more rows
#> 

mtcars |> 
  lazy_dt() |> 
  arrange(dplyr::desc(vs), cyl)
#> Error in `as_string()`:
#> ! Can't convert a call to a string.
#> Backtrace:
#>      ▆
#>   1. ├─dplyr::arrange(lazy_dt(mtcars), dplyr::desc(vs), cyl)
#>   2. └─dtplyr:::arrange.dtplyr_step(...)
#>   3.   └─dtplyr:::capture_dots(.data, ..., .j = FALSE)
#>   4.     └─dtplyr:::map(dots, dt_squash, data = .data, j = .j, is_top = TRUE)
#>   5.       └─base::lapply(.x, .f, ...)
#>   6.         └─dtplyr (local) FUN(X[[i]], ...)
#>   7.           └─dtplyr:::dt_squash(get_expr(x), get_env(x), data, j = j, is_top)
#>   8.             └─dtplyr:::dt_squash_call(x, env, data, j = j)
#>   9.               └─dtplyr:::check_one_arg(x)
#>  10.                 └─rlang::as_name(x[[1]])
#>  11.                   └─rlang::as_string(x)
#>  12.                     └─rlang:::abort_coercion(x, "a string")
#>  13.                       └─rlang::abort(msg, call = call)
```
<sup>Created on 2023-03-28 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

markfairbanks:
Closes #427

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,10 @@
 
 * `consecutive_id()` is now mapped to `data.table::rleid()`.
   Note: `rleid()` only accepts vector inputs and cannot be used with data frame inputs.
+  
+## Minor improvements and bug fixes
+
+* Can namespace calls to `desc()` (#427).
 
 # dtplyr 1.3.1
 
--- a/R/tidyeval.R
+++ b/R/tidyeval.R
@@ -336,8 +336,9 @@ fun_name <- function(fun) {
 }
 
 check_one_arg <- function(x) {
-  fun <- as_name(x[[1]])
-  if (!has_length(x, 2L)) {
+  fun <- call_name(x)
+  args <- call_args(x)
+  if (!has_length(args, 1L)) {
     abort(glue("`{fun}()` expects exactly one argument."))
   }
 }
--- a/tests/testthat/test-tidyeval.R
+++ b/tests/testthat/test-tidyeval.R
@@ -346,6 +346,13 @@ test_that("`desc(col)` is translated to `-col` inside arrange", {
 
   expect_equal(show_query(step), expr(DT[order(-x)]))
   expect_equal(out$x, c("b", "a"))
+
+  # Can namespace `desc()`
+  step <- arrange(dt, dplyr::desc(x))
+  out <- collect(step)
+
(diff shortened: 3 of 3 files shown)


eutwt (approved):
Looks good

PR #432, opened 2023-03-30
Implement `reframe()`

### LINKED ISSUE #417: Implement `reframe()`
This should be easy given that `data.table` aggregation already works just like `reframe()`.

The bigger question - is there anything we can do about restricting `summarize()`? AFAIK this isn't something you can restrict in `data.table`.

``` r
library(data.table)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:data.table':
#> 
#>     between, first, last
#> 
#>     filter, lag
#> 
library(dtplyr)

df <- data.table(x = 1:3, y = c("a", "a", "b"))

df[, .(quantile = quantile(x, c(.25, .5, .75)))]
#>    quantile
#>       <num>
#> 1:      1.5
#> 2:      2.0
#> 3:      2.5

df %>%
  lazy_dt() %>%
  summarize(quantile = quantile(x, c(.25, .5, .75)))
#> Source: local data table [3 x 1]
#> Call:   `_DT1`[, .(quantile = quantile(x, c(0.25, 0.5, 0.75)))]
#> 
#>   quantile
#>      <dbl>
#> 1      1.5
#> 2      2  
#> 3      2.5
#> 
```

markfairbanks:
Closes #417

As of right now this is an alias for `summarize()` that returns an ungrouped lazy_dt. We'll need to make adjustments later when `dplyr` deprecates multi-row returns in `summarize()`


eutwt (approved):
Looks good

PR #449, opened 2023-09-12
Don't squash accessor calls

### LINKED ISSUE #434: rle() returns NULL
I recently updated the dtplyr package and the code snippet with rle() no longer works. The val should have values 19 and 4 but instead both are NULL.

```
library(dplyr)
library(tidyr)
library(dtplyr)

data <- tibble(name = c('a','b'),
             string = c('0000000000000000000','0000hu000000')) %>%
  lazy_dt() %>%
  mutate(val = sapply(string, function(x) rle(strsplit(x, '')[[1]])$lengths[1])) %>%
  collect()
```
image

markfairbanks:
Closes #434

This seemed like the most straightforward solution. There's no way of knowing what objects are being accessed or what the possible name conflicts are.

### DIFF
--- a/R/tidyeval.R
+++ b/R/tidyeval.R
@@ -270,7 +270,7 @@ dt_squash_call <- function(x, env, data, j = TRUE) {
     x[[1]] <- expr(rleid)
     x[-1] <- lapply(x[-1], dt_squash, env, data, j = j)
     x
-  } else if (is_call(x, "$") && !is_mask_pronoun(x)) {
+  } else if (is_call(x, c("$", "[[")) && !is_mask_pronoun(x)) {
     x
   } else {
     x[-1] <- lapply(x[-1], dt_squash, env, data, j = j)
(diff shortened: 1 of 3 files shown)


R/tidyeval.R
```diff
@@ -270,6 +270,8 @@ dt_squash_call <- function(x, env, data, j = TRUE) {
     x[[1]] <- expr(rleid)
     x[-1] <- lapply(x[-1], dt_squash, env, data, j = j)
     x
+  } else if (is_call(x, c("$", "[[")) && !is_mask_pronoun(x)) {
```
eutwt, line 273: I don't think `&& !is_mask_pronoun(x)` has an effect here, since this is an `else if` after `if (is_mask_pronoun(x))`. Looks good either way, that's just a code style preference (maybe you wanted it to be more explicit?)

eutwt (approved):

R/tidyeval.R
markfairbanks: Ah yep good catch. I'll drop that.

